PR #35, opened 2018-03-19
loo_cv() spell check

DavisVaughan:
Remove that `ass`.

It also trimmed some new line characters, I think. It happened automatically.


topepo:
> Remove that ass.

At least I didn't make this typo with the `tidyposterior` package :-)

PR #57, opened 2018-09-10
Rename fill() -> populate() to avoid a conflict with tidyr::fill().

### LINKED ISSUE #49: Consider renaming `fill()` to not clash with `tidyr::fill()`
Currently, loading `rsample` masks `fill()`. Is there a better name?

``` r
library(rsample)
#> Warning: package 'rsample' was built under R version 3.4.4
#> Loading required package: broom
#> Warning: package 'broom' was built under R version 3.4.4
#> Loading required package: tidyr
#> Warning: package 'tidyr' was built under R version 3.4.4
#> 
#> Attaching package: 'rsample'
#> The following object is masked from 'package:tidyr':
#> 
#>     fill
library(tidyr)

mtcars <- mtcars[1:5,]
mtcars$mpg[1] <- NA

fill(mtcars, mpg, .direction = "up")
#> Error in UseMethod("fill"): no applicable method for 'fill' applied to an object of class "data.frame"

tidyr::fill(mtcars, mpg, .direction = "up")
#>                    mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> Mazda RX4         21.0   6  160 110 3.90 2.620 16.46  0  1    4    4
#> Mazda RX4 Wag     21.0   6  160 110 3.90 2.875 17.02  0  1    4    4
#> Datsun 710        22.8   4  108  93 3.85 2.320 18.61  1  1    4    1
#> Hornet 4 Drive    21.4   6  258 110 3.08 3.215 19.44  1  0    3    1
#> Hornet Sportabout 18.7   8  360 175 3.15 3.440 17.02  0  0    3    2
```

Created on 2018-08-15 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

DavisVaughan:
This is a breaking change. @topepo, do you want me to deprecate `fill()` rather than just removing it?

This PR:
- Renames `fill()` to be called `populate()`
- Updates any cross linking of documentation to `fill()` to now point to `populate()`
- Updates NEWS

Closes #49


topepo:
Huh. We're not changing the data structure or names, just a method. Would code using previous `rsample` versions fail (or old objects in the new version) ?

DavisVaughan:
If someone with an old version of `rsample` that used `rsample::fill()` in their code were to update to the new version, their code would fail. They would just need to switch out `rsample::fill()` for `rsample::populate()`

DavisVaughan:
Note to future self) We decided to keep fill for now and deprecate it. The default method of `fill()` will throw the same deprecation warning but call the `tidyr::fill()` function.

PR #58, opened 2018-09-11
Initial time split

### LINKED ISSUE #53: Add `shuffle` arg to initial_split
I'm a big fan of `initial_split()`; feels very natural to make one "big" split, then e.g. k-fold the training set of the big split.

Would it make sense to allow for ordered/non-shuffling samples, for use in timeseries problems where further sampling would be done by `rolling_origin`?

The easiest solution is probably to add an arg like `shuffle = FALSE` to `initial_split`, which would mirror how `scikit-learn` does it (http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html).

It's easy to imagine going deeper, allowing for e.g. the `skip` arg of `rolling_origin`, or an `order_by` (similar to https://github.com/tidymodels/recipes/issues/171), but starting small with a `shuffle` arg might be a good start.

ClaytonJY:
Closes #53.
Best if squashed.

The new function is dead simple, even using `rsplit` instead of `make_split` because that's more straightforward and unlike `initial_split` there's no obvious additional class to assign (though perhaps one should be added?). Skipped `strata` because it's not clear if/how to do that, and also because I dream about replacing that arg with `group_by`.

Tests mirror those for `initial_split` as closely as possible; one additional line to check for order.

NEWS is updated, but updating `pkgdown` didn't go so well; seems `tidytemplate` is missing from Suggests + Remote, bunch of things were getting updated due to version changes, and the `Keras` vignette was taking forever.


R/initial_split.R
```diff
@@ -1,24 +1,25 @@
 #' Simple Training/Test Set Splitting
 #'
-#' `initial_split` creates a single binary split of the data 
-#'  into a training set and testing set. `training` and
-#'  `testing` are used to extract the resulting data.
+#' `initial_split` creates a single binary split of the data into a training set
+#' and testing set. `initial_split` does the same, but takes the _first_ `prop`
```
topepo: `initial_time_split` does the same.

NEWS.md
```diff
@@ -1,7 +1,8 @@
 # `rsample` 0.0.2.9000
 
 * Changed the R version requirement to be R >= 3.1 instead of 3.3.3. 
-* The `recipes`-related `prepper` function was [moved to the `recipes` package](https://github.com/tidymodels/rsample/issues/48). This makes the `rsample` install footprint much smaller. 
+* The `recipes`-related `prepper` function was [moved to the `recipes` package](https://github.com/tidymodels/rsample/issues/48). This makes the `rsample` install footprint much smaller.
+* Added function `initial_time_split` for ordered initial sampling appropriate for timeseries data.
```
topepo: "time series"

R/initial_split.R
```diff
@@ -1,24 +1,25 @@
...
+#' @param strata A variable that is used to conduct stratified sampling to create the resamples.
 #' @export
-#' @return  An `rset` object that can be used with the `training` and `testing` functions to extract the data in each split. 
+#' @return  An `rset` object that can be used with the `training` and `testing` functions to extract the data in each split.
 #' @examples
 #' set.seed(1353)
 #' car_split <- initial_split(mtcars)
 #' train_data <- training(car_split)
-#' test_data <- testing(car_split) 
+#' test_data <- testing(car_split)
```
topepo: Maybe show an example with the `drinks` data?

ClaytonJY:
Changes made, including the new example.

Is the slight decrease in code-coverage an issue? I could add a test that hits the validation of `prop` to remedy that, I think.

ClaytonJY:
Updated relevant `pkgdown` pages, just two of them.

topepo:
Looks fine to me. 

You can get the css at `https://github.com/tidyverse/tidytemplate`. If that is installed, it works fine. 

I don't included it in Suggests since `R CMD check` or `build` don't rely on it. I'll update the pkgdown files to make a note on its location.

Issue #37, opened 2018-05-19
Allow for resampling of vectors


RishiSadhir:
Max - Thanks for all of the excellent work you do!

I was writing some code to calculate bootstrapped statistics and found the desire to do this on a vector instead of a dataframe. 

Cheers.

### Minimal, reproducible example:
``` r
# Setup
library(rsample)
library(purrr)

# What I want to accomplish
statistic <- rnorm(1000)
bootstrapped_mean <- map_dbl(bootstraps(statistic)$splits, ~ mean(analysis(.x)))
#> Error: `num` should be > 0
summary(bootstrapped_mean)
#> Error in summary(bootstrapped_mean): object 'bootstrapped_mean' not found

# What I have to do
df = data.frame(statistic = rnorm(1000))
bootstrapped_mean <- map_dbl(bootstraps(df)$splits, ~ mean(analysis(.x)$statistic))
summary(bootstrapped_mean)
#>     Min.  1st Qu.   Median     Mean  3rd Qu.     Max. 
#> -0.10787 -0.09432 -0.06594 -0.06679 -0.04551  0.04240
```

<details>
<summary>Session info</summary>

``` r
devtools::session_info()
#> Session info -------------------------------------------------------------
#>  setting  value                       
#>  version  R version 3.5.0 (2018-04-23)
#>  system   x86_64, mingw32             
#>  ui       RTerm                       
#>  language (EN)                        
#>  collate  English_United States.1252  
#>  tz       America/New_York            
#>  date     2018-05-19
#> Packages -----------------------------------------------------------------
#>  package    * version    date       source        
#>  abind        1.4-5      2016-07-21 CRAN (R 3.5.0)
#>  assertthat   0.2.0      2017-04-11 CRAN (R 3.5.0)
#>  backports    1.1.2      2017-12-13 CRAN (R 3.5.0)
#>  base       * 3.5.0      2018-04-23 local         
#>  bindr        0.1.1      2018-03-13 CRAN (R 3.5.0)
#>  bindrcpp     0.2.2      2018-03-29 CRAN (R 3.5.0)
#>  broom      * 0.4.4      2018-03-29 CRAN (R 3.5.0)
#>  class        7.3-14     2015-08-30 CRAN (R 3.5.0)
#>  compiler     3.5.0      2018-04-23 local         
#>  CVST         0.2-1      2013-12-10 CRAN (R 3.5.0)
#>  datasets   * 3.5.0      2018-04-23 local         
[...]

PR #71, opened 2018-11-22
correctly count n for vectors

l4u532:
when supplying a vector (e.g. `c(1,2,3,4)`), usually the following error gets thrown:

`There should be at least ",
         initial + assess,
         " nrows in `data`"`

this because `nrow()` instead of `NROW()` is used. The former does not support atomic vectors.

### DIFF
--- a/R/rolling_origin.R
+++ b/R/rolling_origin.R
@@ -37,7 +37,11 @@
 #' @export
 rolling_origin <- function(data, initial = 5, assess = 1,
                            cumulative = TRUE, skip = 0, ...) {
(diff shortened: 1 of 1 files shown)


DavisVaughan:
Hi @l4u532, we actually make no guarantees that anything besides a data frame will work with the functions in `rsample`. The `data` argument to `rolling_origin()` even says "A data frame". We don't feel that it is worth it to support vectors, but we appreciate your interest and thanks for submitting the PR!

See also this issue:
https://github.com/tidymodels/rsample/issues/37

Issue #96, opened 2019-06-25
Is online documentation out of date? (versions of package do not match)


GegznaV:
Why do versions of some **tidymodels** and **tidyverse** packages and their documentation do not match? This time I found this discrepancy in `rsample` package, e.g.:
![image](https://user-images.githubusercontent.com/12725868/60126910-abd17e80-9797-11e9-8464-933fbd0a6421.png)

Is documentation out of date or there are some other issues (e.g., issues relatied to the automatic deployment of documentation website, Travis CI, etc.)?

topepo:
I haven't updated it. I'll do that in the next day or so.

PR #98, opened 2019-07-03
confidence intervals for bootstrap estimates 

topepo:
Three functions to create different types of intervals.

`pkgdown` sites were also changed to update and fixed some bugs.


topepo:
also closes #96

PR #101, opened 2019-07-12
Breaks feature

### LINKED ISSUE #77: Change the strata groups for quantitative response in initial_split()
From the [Applied Machine Learning Workshop](https://github.com/topepo/rstudio-conf-2019) at rstudio::conf.

I asked how the initial_split function calculates strata for quantitative response variables. Answer: it splits by quartiles. I asked about changing this to e.g. deciles, and you told me to file an issue. 

Here's the feature request: add capability to split by deciles, etc. when response being stratified is quantitative. 

Thanks.

jyuu:
Small addition to the `make_strata` function: 

```{r} 
    if (floor(n / breaks) < depth) {
      warning(paste0("The number of observations in each quantile is ",
              "below the recommended threshold of ", depth, ". Stratification ",
              "will be done with ", floor(n/depth), " breaks instead."),
              call. = FALSE)
    }
    breaks <- min(breaks, floor(n/depth))
```

Changes to any functions that utilize `make_strata` is also made to include the `breaks` flag 

Closes #77

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * Added three functions to compute different bootstrap confidence intervals. 
 * A new function (`add_resample_id`) augments a data frame with columns for the resampling identifier. 
+* Updated `initial_split`, `mc_cv`, `vfold_cv`, `bootstraps`, and `group_vfold_cv` to use tidyselect on the stratification variable.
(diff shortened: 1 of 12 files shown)


topepo:
Looks great. Can you update the NEW.md file to describe this change (and the `tidyselect` changes from the last PR)?

topepo:
Thanks again

PR #104, opened 2019-07-12
Confint api test number 2000

fbchow:
Match broom::tidy() column name


topepo:
Just in time.. I'm prepping for a CRAN release. Thanks!

PR #114, opened 2019-11-13
New unnest interface in tidyr 1.0

fbchow:
Noticed new `unnest` interface from tidyverse/tidyr#702 and tidyverse/tidyr@953175f warning message


DavisVaughan:
Thanks!

PR #132, opened 2020-02-21
Add `pool` argument to vfold_cv stratified sampling

### LINKED ISSUE #110: Modify strata_check defaults from vfold_cv
I have a dataset with seven classes, and 20 observations in each class. I want to do three-fold cross-validation on this dataset, using stratified sampling, because when using unstratified sampling there is a chance that a particular fold will not contain any positive examples for a given class. 

However, when I try to do this with `vfold_cv` I get the following warning:

``` r
library(tidyverse)
library(rsample)

X = matrix(rnorm(140 * 100), ncol = 100, nrow = 140)
y = rep(letters[1:7], each = 20)
dat = as.data.frame(X) %>%
  mutate(label = y)
cv = vfold_cv(dat, v = 3, strata = 'label')
#> Warning: Too little data to stratify. Unstratified resampling will be used.
```

<sup>Created on 2019-09-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

It seems this is related to the `check_strata` function, specifically the default value `pool = 0.15`. In the context of `check_strata`, `pcts` is a vector of length 7 where every value is equal to 1 / 7 and so the function returns a single stratum:

```
    num_vals <- unique(x)
    n <- length(x)
    num_miss <- sum(is.na(x))
    if (length(num_vals) <= nunique | is.character(x) | is.factor(x)) {
        x <- factor(x)
        xtab <- sort(table(x))
        pcts <- xtab/n
        if (all(pcts < pool)) {
            warning("Too little data to stratify. Unstratified resampling ", 
                "will be used.", call. = FALSE)
            return(factor(rep("strata1", n)))
        }
```

This wouldn't be an issue if I could just change the default value of `pool` from `vfold_cv` but at present it doesn't seem like I can. Is it possible to pass the ellipsis from `vfold_cv`->`vfold_splits`->`make_strata`? 

sessionInfo():
```
R version 3.5.2 (2018-12-20)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS High Sierra 10.13.6

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRlapack.dylib

locale:
[1] en_CA.UTF-8/en_CA.UTF-8/en_CA.UTF-8/C/en_CA.UTF-8/en_CA.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] rsample_0.0.5     forcats_0.4.0     stringr_1.4.0     dplyr_0.8.3       purrr_0.3.2       readr_1.3.1       tidyr_0.8.99.9000
 [8] tibble_2.1.3      ggplot2_3.2.1     tidyverse_1.2.1  

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.2       cellranger_1.1.0 pillar_1.4.2     compiler_3.5.2   tools_3.5.2      digest_0.6.20    zeallot_0.1.0    jsonlite_1.6    
 [9] lubridate_1.7.4  lifecycle_0.1.0  nlme_3.1-141     gtable_0.3.0     lattice_0.20-38  pkgconfig_2.0.2  rlang_0.4.0      cli_1.1.0       
[17] rstudioapi_0.10  parallel_3.5.2   yaml_2.2.0       haven_2.1.1      furrr_0.1.0      withr_2.1.2      xml2_1.2.2       httr_1.4.1      
[25] globals_0.12.4   generics_0.0.2   vctrs_0.2.0      hms_0.5.1        grid_3.5.2       tidyselect_0.2.5 glue_1.3.1       listenv_0.7.0   
[33] R6_2.4.0         readxl_1.3.1     modelr_0.1.5     magrittr_1.5     codetools_0.2-16 backports_1.1.4  scales_1.0.0     rvest_0.3.4     
[41] assertthat_0.2.1 future_1.14.0    colorspace_1.4-1 stringi_1.4.3    lazyeval_0.2.2   munsell_0.5.0    broom_0.5.2      crayon_1.3.4    
```

swt30:
If you attempt to sample from strata that don't have enough examples, `vfold_cv`  bails out with a 'too little data to stratify' message and falls back to unstratified sampling. Sometimes we might want to force it to continue anyway – an example is given in #110. I also use this function on some unbalanced strata and want to ensure that a minimum of one example from each is included, even at the cost of some representativeness.

This PR achieves that by passing the `pool` argument through from `vfold_cv` to `make_strata`. Then you can use `vfold_cv(...,  pool = 0)` to force stratificiation to occur. The default value of `pool` is 0.15 which matches the value in `make_strata`.

Closes #110

### DIFF
--- a/R/vfold.R
+++ b/R/vfold.R
@@ -20,8 +20,7 @@
 #' @param strata A variable that is used to conduct stratified sampling to
 #'  create the folds. This could be a single character value or a variable name
 #'  that corresponds to a variable that exists in the data frame.
-#' @param breaks A single number giving the number of bins desired to stratify
-#'  a numeric stratification variable.
+#' @inheritParams make_strata
 #' @param ... Not currently used.
 #' @export
 #' @return A tibble with classes `vfold_cv`, `rset`, `tbl_df`, `tbl`, and
@@ -62,7 +61,8 @@
 #'           mean(dat == "virginica")
 #'         })
 #' @export
-vfold_cv <- function(data, v = 10, repeats = 1, strata = NULL, breaks = 4, ...) {
+vfold_cv <- function(data, v = 10, repeats = 1, strata = NULL, breaks = 4,
+                     pool = 0.15, ...) {
 
   if(!missing(strata)) {
     strata <- tidyselect::vars_select(names(data), !!enquo(strata))
@@ -72,10 +72,11 @@ vfold_cv <- function(data, v = 10, repeats = 1, strata = NULL, breaks = 4, ...) 
   strata_check(strata, names(data))
 
   if (repeats == 1) {
-    split_objs <- vfold_splits(data = data, v = v, strata = strata, breaks = breaks)
+    split_objs <- vfold_splits(data = data, v = v, strata = strata,
+                               breaks = breaks, pool = pool)
   } else {
     for (i in 1:repeats) {
-      tmp <- vfold_splits(data = data, v = v, strata = strata)
+      tmp <- vfold_splits(data = data, v = v, strata = strata, pool = pool)
       tmp$id2 <- tmp$id
       tmp$id <- names0(repeats, "Repeat")[i]
       split_objs <- if (i == 1)
@@ -109,7 +110,7 @@ vfold_complement <- function(ind, n) {
 #' @importFrom tibble tibble
 #' @importFrom purrr map
 #' @importFrom dplyr bind_rows
-vfold_splits <- function(data, v = 10, strata = NULL, breaks = 4) {
+vfold_splits <- function(data, v = 10, strata = NULL, breaks = 4, pool = 0.15) {
   if (!is.numeric(v) || length(v) != 1)
     stop("`v` must be a single integer.", call. = FALSE)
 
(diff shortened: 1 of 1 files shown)


topepo:
See [my comments](https://github.com/tidymodels/rsample/issues/110#issuecomment-591714120)

juliasilge:
Thank you so much for this PR @swt30! 🙌 We are going to close it in favor of lowering the threshold a bit, because of past experience with exposing this kind of argument. We would be happy to see a PR for that change, if you are interested.

If you'd like to check out guidelines for contributing in the future, please [check out our new page on `tidymodels.org`](https://www.tidymodels.org/contribute/).

swt30:
Thanks @juliasilge – have moved on from this problem now but I will be sure to give the lowered threshold a try if I have to do something similar in the future.

juliasilge:
FYI we implemented the lowered threshold in #149. 👍

PR #137, opened 2020-03-28
Fix compatibility with tibble 3.0.0

krlmlr:
Columns now can have inner names, the test wasn't prepared for this.

Applies cleanly to the 0.0.5 tag if you want to do a patch release.


topepo:
Already done in #134

PR #146, opened 2020-04-02
Update interpretation of confidence interval in Working_with_rsets.Rmd

### LINKED ISSUE #145: Vignette - interpretation of confidence interval
In my opinion, the following sentence of the vignette (Working_with_rsets) is a bit misleading: 

> _On average_, there is no evidence for a difference in the genders.

I would suggest something like the following: 

> The calculated 95% confidence interval contains 0. Hence, there are no significant differences in the median monthly income between genders at a confidence level of 95%.  

Also, I would exclude the 50% quantile in

`quantile(bt_resamples$wage_diff, probs = c(0.025, 0.500, 0.975))`

because it could be mistaken for the usual point estimate, and I would say a confidence interval is usually defined by two values.

NikolaiVogl:
Closes #145

### DIFF
--- a/vignettes/Working_with_rsets.Rmd
+++ b/vignettes/Working_with_rsets.Rmd
@@ -171,7 +171,7 @@ quantile(bt_resamples$wage_diff,
          probs = c(0.025, 0.975))
 ```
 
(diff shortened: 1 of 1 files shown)


juliasilge:
GH actions failed because recipes is not available on macOS

juliasilge:
Thanks for the PR @NikolaiVogl! 🙌

PR #150, opened 2020-05-07
Add vctrs support

DavisVaughan:
Closes #116 

This PR adds vctrs support to rsample. It is broken into 3 sections:

- Base R methods that have been added for supporting dplyr

- vctrs methods

- dplyr specific methods

It also adds an enormous test suite to ensure that the expected dplyr/vctrs behavior is happening for _all_ rset subclasses.

---

Base:

- `[.rset` was rewritten from the ground up

- `names<-.rset` was added

- I also had to special case `nested_cv`, so `[.nested_cv` and `names<-.nested_cv` have also been added.

All of the above generally take the approach of dropping back to a bare tibble any time any of the rset specific columns are modified or dropped in any way.

---

For vctrs:

- It looks like a lot of methods, but really they all repeat themselves. We have to do it this way because there is no inheritance in vctrs `vec_ptype2()` and `vec_cast()`.

- `vec_restore()` methods were added. These are called right before exiting from `vec_slice()` or `vec_ptype()` (among other things). The general idea is to always "restore" to a bare tibble rather than trying to restore to the rset subclass. This simplifies things considerably, and is generally the right approach because:

   - When you row subset an rsample subclass, it no longer has the correct structure. i.e. if you slice the first 5 rows of a 10-fold cross-validation rset, it is no longer a valid rset. So it makes more sense to return a tibble.

   - It influences the common types that appear in `vec_ptype2()`.

      - The common type of, say, a `<bootstraps>` with another `<bootstraps>` falls back to a `<tbl_df>`

      - The common type of `<bootstraps>` with `<tbl_df>` also falls back to `<tbl_df>`

      This is correct behavior because it affects things like `vec_c(<bootstraps>, <bootstraps>)`. Like with row subsetting, you probably won't have a valid bootstrap data frame after this (you'll have too many rows, and who knows if those two bootstrap data frames were even compatible), so the right thing to do is to fall back to returning a tibble.

- `vec_ptype2()` methods were added, but they abort immediately. These aren't strictly required, but I wanted to ensure that they really are never called because this is sort of an experimental idea. They aren't required because by having `vec_restore()` restore to a bare tibble, `vec_ptype2()` will never get the chance to dispatch off of `<bootstraps>`, so it should never see these methods.

- `vec_cast()` methods were added.

   - The following error. They will never be called "automatically" (because of how `vec_restore()` works), and would only ever show up if a user explicitly called `vec_cast()` directly. This is unlikely, and casting to a rset-subclass is essentially undefined, which is why these error.

     ```
     vec_cast(<rset-subclass>, <rset-subclass>) 
     vec_cast(<tbl_df>, <rset-subclass>)
     vec_cast(<data.frame>, <rset-subclass>)
     ```

   - The following are allowed. In things like `vec_c(<bootstraps>, <tbl_df>)`, when `vec_ptype2()` returns a bare tibble as the common type the `vec_cast(to = <tbl_df>)` method is used to convert the rset subclass to a bare tibble.

     ```
     vec_cast(<rset-subclass>, <tbl_df>)
     vec_cast(<rset-subclass>, <data.frame>)
     ```

--- 

For dplyr:

- `dplyr_reconstruct.rset()` works exactly like `vec_restore()`. Rather than restoring to a rset subclass after an operation like a row slice, filter, or left-join, we instead restore to a bare tibble because it is highly unlikely that we still have a valid rset subclass.

- `dplyr_col_modify.rset()` calls the default col-modify method from dplyr, then checks if we modified any rset specific columns. If we did, then we have to fall back to a bare tibble because the user has likely altered the column in a way that makes it no longer a valid rset subclass. Otherwise, we keep the rset subclass.

- `dplyr_row_slice.rset()` is not required because the default behavior is fine. It calls `vec_slice()` and then hands off to `dplyr_reconstruct()`, which strips the rset subclass.

Implications of these methods are that most dplyr verbs will return a bare tibble when used with an rsample subclass, but there are a few key ones that can retain the rset subclass:

- `mutate()` when the columns modified don't touch rset specific columns (`splits`, `id`)

- `select()` when the user retains the rset specific columns

- `relocate()`

- `rename()` when the renamed column isn't an rset specific column. Additionally if a column that starts with `id` is renamed, then as long as it still starts with `id` and doesn't duplicate any other column, it is still a valid rset.


topepo:
>  you slice the first 5 rows of a 10-fold cross-validation rset, it is no longer a valid rset.

I might not want to die on this hill just yet but we should revisit this down the line. I'm opposed to expanding a tibble since the id fields are not longer unique but subsetting doesn't feel disqualifying (to me). 

We will also have some applications where we do that. However we could get around this constraint in other ways (e.g. keep the whole `rset` but don't populate some rows for non-essential columns.

topepo:
The join behavior is unexpected to me. In all cases, it is downcast to a tibble. I would not expect that to occur with `full_join()` and left/right joins where the `rset` is not changed.  

``` r
library(tidymodels)
#> ── Attaching packages ──────────────────────────────────────────── tidymodels 0.1.0 ──
#> ✓ broom     0.5.4           ✓ recipes   0.1.12     
#> ✓ dials     0.0.6           ✓ rsample   0.0.6.9000 
#> ✓ dplyr     0.8.99.9002     ✓ tibble    3.0.1      
#> ✓ ggplot2   3.3.0           ✓ tune      0.1.0      
#> ✓ infer     0.5.1           ✓ workflows 0.1.0      
#> ✓ parsnip   0.1.1           ✓ yardstick 0.0.5      
#> ✓ purrr     0.3.4
#> ── Conflicts ─────────────────────────────────────────────── tidymodels_conflicts() ──
#> x purrr::discard()  masks scales::discard()
#> x dplyr::filter()   masks stats::filter()
#> x dplyr::lag()      masks stats::lag()
#> x ggplot2::margin() masks dials::margin()
#> x recipes::step()   masks stats::step()

set.seed(33)
rs <- vfold_cv(mtcars)

new_rows <- tibble(id = paste0("Fold0", 1:3), potato = 1:3)

# Makses sense since we lose rows of `rs`:
anti_join(rs, new_rows, by = "id")
#> # A tibble: 7 x 2
#>   splits         id    
#>   <named list>   <chr> 
#> 1 <split [29/3]> Fold04
#> 2 <split [29/3]> Fold05
#> 3 <split [29/3]> Fold06
#> 4 <split [29/3]> Fold07
#> 5 <split [29/3]> Fold08
#> 6 <split [29/3]> Fold09
#> 7 <split [29/3]> Fold10
right_join(rs, new_rows, by = "id")
#> # A tibble: 3 x 3
#>   splits         id     potato
#>   <named list>   <chr>   <int>
#> 1 <split [28/4]> Fold01      1
#> 2 <split [28/4]> Fold02      2
#> 3 <split [29/3]> Fold03      3

# Maybe not makes sense since `rs` is still intact
full_join(rs,  new_rows, by = "id")
#> # A tibble: 10 x 3
#>    splits         id     potato
#>    <named list>   <chr>   <int>
#>  1 <split [28/4]> Fold01      1
#>  2 <split [28/4]> Fold02      2
#>  3 <split [29/3]> Fold03      3
#>  4 <split [29/3]> Fold04     NA
#>  5 <split [29/3]> Fold05     NA
#>  6 <split [29/3]> Fold06     NA
#>  7 <split [29/3]> Fold07     NA
#>  8 <split [29/3]> Fold08     NA
#>  9 <split [29/3]> Fold09     NA
#> 10 <split [29/3]> Fold10     NA
left_join(rs, new_rows, by = "id")
#> # A tibble: 10 x 3
#>    splits         id     potato
#>    <named list>   <chr>   <int>
#>  1 <split [28/4]> Fold01      1
#>  2 <split [28/4]> Fold02      2
#>  3 <split [29/3]> Fold03      3
#>  4 <split [29/3]> Fold04     NA
#>  5 <split [29/3]> Fold05     NA
#>  6 <split [29/3]> Fold06     NA
#>  7 <split [29/3]> Fold07     NA
#>  8 <split [29/3]> Fold08     NA
#>  9 <split [29/3]> Fold09     NA
#> 10 <split [29/3]> Fold10     NA
```

<sup>Created on 2020-05-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
The reason that the joins always fall back to a tibble is that it is nearly impossible to always be able to tell if you still have a valid rset subclass when just comparing the original `x` with the result of the join.

In the case of a `right_join()`, for example, you can end up with a result that:
- Still has correctly named columns
- Still has the same number of rows as `x`
- Is not a valid bootstraps tibble

```r
library(dplyr, warn.conflicts = FALSE)
library(rsample)

df1 <- bootstraps(mtcars, 2)
df1
#> # Bootstrap sampling 
#> # A tibble: 2 x 2
#>   splits          id        
#>   <list>          <chr>     
#> 1 <split [32/11]> Bootstrap1
#> 2 <split [32/13]> Bootstrap2

df2 <- data.frame(id = c("Bootstrap1", "Bootstrap1"))

# number of rows in x is still the same
dplyr::right_join(df1, df2, by = "id")
#> # A tibble: 2 x 2
#>   splits          id        
#>   <list>          <chr>     
#> 1 <split [32/11]> Bootstrap1
#> 2 <split [32/11]> Bootstrap1
```

Because it is so difficult, I think it is better to consistently return a tibble.

topepo:
That will break some existing code. We're also going to want to do this with model stacking. 

There might be a minimum that we can do: 

``` r
library(tidymodels)
#> ── Attaching packages ──────────────────────────────────────────── tidymodels 0.1.0 ──
#> ✓ broom     0.5.4           ✓ recipes   0.1.12     
#> ✓ dials     0.0.6           ✓ rsample   0.0.6      
#> ✓ dplyr     0.8.99.9002     ✓ tibble    3.0.1      
#> ✓ ggplot2   3.3.0           ✓ tune      0.1.0      
#> ✓ infer     0.5.1           ✓ workflows 0.1.0      
#> ✓ parsnip   0.1.1           ✓ yardstick 0.0.5      
#> ✓ purrr     0.3.4
#> Warning: package 'rsample' was built under R version 3.6.2
#> ── Conflicts ─────────────────────────────────────────────── tidymodels_conflicts() ──
#> x purrr::discard()  masks scales::discard()
#> x dplyr::filter()   masks stats::filter()
#> x dplyr::lag()      masks stats::lag()
#> x ggplot2::margin() masks dials::margin()
#> x recipes::step()   masks stats::step()

set.seed(33)
rs <- vfold_cv(mtcars)

new_rows <- tibble(id = paste0("Fold0", 1:3), potato = 1:3)

check_still_rset <- function(pre, post) {
  isTRUE(
    all.equal(
      pre  %>% select(splits, starts_with("id")), 
      post %>% select(splits, starts_with("id")),
      check.attributes = FALSE 
    )
  )
}

check_still_rset(rs,  full_join(rs,  new_rows, by = "id"))
#> [1] TRUE
check_still_rset(rs, inner_join(rs,  new_rows, by = "id"))
#> [1] FALSE
```

<sup>Created on 2020-05-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

topepo:
Also, are we requiring the new `dplyr` for tidymodels packages?

juliasilge:
This is a really good question:

> Also, are we requiring the new dplyr for tidymodels packages?

I tried installing this to test things out but couldn't quite figure out the answer.

What are the possibilities here? Is it possible for, say, rsample to work for both pre- and post-dplyr 1.0.0?

DavisVaughan:
Okay we now have support for dplyr < 1.0.0. I am requiring dplyr 0.8.5 though. I think that seems fair, but we can talk about that too.

The "switches" that get turned on / off depending on the dplyr version are in 3 places:

- `compat-dplyr.R` holds the dplyr 1.0.0 compat info. `compat-dplyr-old.R` holds the pre dplyr 1.0.0 compat info. These are conditionally run based on the dplyr version. The way this is set up, on dplyr 1.0.0 you shouldn't see `arrange.rset` in the package whatsoever, even `rsample:::arrange.rset` won't bring it up.

- In addition to the above, `zzz.R` conditionally registers the S3 methods in the compat-dplyr files depending on the dplyr version.

- Lastly, there is a `test-compat-dplyr.R` and a `test-compat-dplyr-old.R`.
   - The tests in `test-compat-dplyr.R` are only run with dplyr 1.0.0 and above
   - The tests in `test-compat-dplyr-old.R` are run on all dplyr versions

The testing setup is done so we ensure that the 6 dplyr verbs that were supported pre-dplyr-1.0.0 work the same way as on dplyr 1.0.0. When users upgrade to dplyr 1.0.0, they get additional support for all other dplyr verbs.

There is a small help page under ``` ?`rsample-dplyr` ``` that talks about the compat support.

topepo:
Not sure where to put this (maybe belongs in a blog post): 

````

#' These examples are not exhaustive and may not capture some edge-cases.
#'
#' ------------------------------------------------------------------------------
#' Joins
#'
#' Joins that reduce rows of the original rset object (i.e. nrow(tbl) < nrow(rset))
#'
#'                          old rsample/old dplyr | new rsample old dplyr | new rsample new dplyr
#' anti_join(rset, tbl)     rset                  | rset                  | tibble
#' full_join(rset, tbl)     error                 | error                 | rset
#' inner_join(rset, tbl)    error                 | error                 | tibble
#'
#' The idea here is that, if there are less rows in the result, the result should
#' not be an rset object. For example, you can't have a 10-fold CV object
#' without 10 rows.
#'
#' Joins that keep the rows of the original rset object (i.e. nrow(tbl) == nrow(rset))
#'
#'                          old rsample/old dplyr | new rsample old dplyr | new rsample new dplyr
#' full_join(rset, tbl)     error                 | error                 | rset
#' left_join(rset, tbl)     error                 | error                 | rset
#'
#' As with the logic above, if the orignal rset object (defined by the split
#' column and the id column(s)) is left intact, the results should be an rset.
#'
#' ------------------------------------------------------------------------------
#' Row subsetting
#'
#' As mentioned above, this should result in a tbl if any rows are removed.
#'
#'                          old rsample/old dplyr | new rsample old dplyr | new rsample new dplyr
#' rset_obj[ind,]           tibble                | tibble                | tibble
#' rset_obj %>% slice()     rset                  | tibble                | tibble
#' rset_obj %>% filter()    rset                  | tibble                | tibble
#' rset_obj %>% arrange()   rset                  | rset                  | rset
#'
#' In the case of `arrange()` row order doesn't matter as long as the orginal
#' rows are present.
#'
#' ------------------------------------------------------------------------------
#' Column subsetting
#'
#' Cases when the required columns are removed or renamed:
#'
#'                          old rsample/old dplyr | new rsample old dplyr | new rsample new dplyr
#' rset_obj[,ind]           tibble                | tibble                | tibble
#' rset_obj %>% select()    rset                  | tibble                | tibble
#' rset_obj %>% rename()    rset                  | tibble                | tibble
#'
#'
#' Cases when no required columns are affected
#'
#'                          old rsample/old dplyr | new rsample old dplyr | new rsample new dplyr
#' rset_obj[,ind]           tibble                | rset                  | rset
#' rset_obj %>% select()    rset                  | rset                  | rset
#'
#' ------------------------------------------------------------------------------
#' Other column operations
#'
#' Cases when no required columns are affected
#'
#'                          old rsample/old dplyr | new rsample old dplyr | new rsample new dplyr
#' rset_obj %>% mutate()    rset                  | rset                  | rset
````

**Comment by juliasilge** [comment]
This is pretty helpful seeing it laid out like this, at least for me.

I was just looking at the new tests and that is a *lot* of this that is being covered there. Am I understanding correctly that one test file is for tests that we plan to pass on old + new dplyr, and one test file only passes on new dplyr? Are we going to install old dplyr in CI somehow?

File: R/zzz.R [hunk 1]
```diff
@@ -0,0 +1,28 @@
+# nocov start
+
+# From vctrs
+on_package_load <- function(pkg, expr) {
```
topepo, line 4: That's fancy. Why is that required?

R/rset.R
```diff
@@ -0,0 +1,273 @@
...
+  }
+}
+
+# ------------------------------------------------------------------------------
+
+col_equals_splits <- function(x) {
+  vec_equal(x, "splits")
+}
+
+col_starts_with_id <- function(x) {
```
topepo, line 267: This might be better:

```r
vals <- c("id", "id2", "id3", "id_val", "id2_thingy")
grepl("(^id$)|(^id[1-9]$)", vals)
```

DavisVaughan:
> I was just looking at the new tests and that is a lot of this that is being covered there

@juliasilge the goal is to be exhaustive in testing dplyr verbs, over _all_ rset subclasses. Which is why each test is really a loop over each type of rset subclass.

> one test file is for tests that we plan to pass on old + new dplyr, and one test file only passes on new dplyr

That is correct. `test-dplyr-compat-old.R` runs on both old and new dplyr. `test-dplyr-compat.R` runs on only new dplyr. When we eventually deprecate support for old dplyr, the tests in `test-dplyr-compat-old.R` should move to `test-dplyr-compat.R`.

> Are we going to install old dplyr in CI somehow?

I already install dev dplyr in CI for one of the Mac builds, the rest of the builds use the current dplyr version. When dplyr 1.0.0 is released, I will switch the CI to have a special build that installs dplyr 0.8.5 with `remotes::install_version()`.

R/compat-vctrs.R
```diff
@@ -0,0 +1,560 @@
...
+# So when `vec_ptype2()` is called on an rset, that rset is downgraded to a
+# bare tibble _before_ the search for a `vec_ptype2()` method occurs. This
+# means that it will never find a ptype2 method such as
+# `vec_ptype2.bootstraps.tbl_df()`, because the <bootstraps> will become a
+# tbl_df first. This means that we rely entirely on the `tbl_df` ptype2 methods,
+# which are already implemented in vctrs to return another tbl_df.
+#
+# The implications of this are that whenever a rset subclass is combined with
+# another rset subclass or bare tibble, the resulting common type is always
+# another bare tibble. So if you `vec_c(<bootstraps>, <tbl_df>)` the result
```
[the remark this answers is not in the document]
DavisVaughan: Really `vec_c()` is just being used as an example of a vctrs function that computes the common type of its inputs. I could have also used `vec_rbind()`. The point is just to show that generally it falls back to tibble when working with vctrs functions

R/zzz.R
DavisVaughan: It isn't, I thought it was. I removed it.

DavisVaughan:
I incorporated @topepo's documentation into the dplyr compat help section. It also shows up in the pkgdown function reference.

(Side note that we really need to get pkgdown working automatically for rsample, and remove the current docs folder if at all possible)

I think they look pretty sleek in R's help and with pkgdown


Screen Shot 2020-05-14 at 10 56 45 AM


Screen Shot 2020-05-14 at 10 56 24 AM

PR #163, opened 2020-06-19
Remove iris dataset

juliasilge:
Remove iris dataset from

- tests
- GitHub issue template
- docs


juliasilge:
I believe this is all of it!

Issue #42, opened 2018-07-10
A idea for an example, rolling_origin() over irregular time chunks


DavisVaughan:
This is not an issue, more of an example idea for `rolling_origin()`.

I've been working closely with the _tsibble_ team on some of their rolling functions, and we worked on rolling over irregular calendar periods. I think this could be a neat addition for `rolling_origin()` and luckily it's already built in. This could be a neat example for the docs somewhere, and I think the idea overall is pretty powerful, especially for time series modeling work where you want to ensure you are using calendar windows rather than fixed windows of say 5 periods.

``` r
library(rsample)
library(riingo)
library(tidyverse)
library(tsibble)

aapl <- riingo::riingo_prices("AAPL", start_date = "2000-01-01")

aapl
#> # A tibble: 4,658 x 14
#>    ticker date                close  high   low  open volume adjClose
#>    <chr>  <dttm>              <dbl> <dbl> <dbl> <dbl>  <int>    <dbl>
#>  1 AAPL   2000-01-03 00:00:00 112.  112.  102.  105.  4.78e6     3.56
#>  2 AAPL   2000-01-04 00:00:00 102.  111.  101.  108.  4.57e6     3.26
#>  3 AAPL   2000-01-05 00:00:00 104   111.  103   104.  6.95e6     3.30
#>  4 AAPL   2000-01-06 00:00:00  95   107    95   106.  6.86e6     3.02
#>  5 AAPL   2000-01-07 00:00:00  99.5 101    95.5  96.5 4.11e6     3.16
#>  6 AAPL   2000-01-10 00:00:00  97.8 102.   94.8 102   4.51e6     3.10
#>  7 AAPL   2000-01-11 00:00:00  92.8  99.4  90.5  95.9 3.94e6     2.95
#>  8 AAPL   2000-01-12 00:00:00  87.2  95.5  86.5  95   8.71e6     2.77
#>  9 AAPL   2000-01-13 00:00:00  96.8  98.8  92.5  94.5 9.22e6     3.07
#> 10 AAPL   2000-01-14 00:00:00 100.  102.   99.4 100   3.49e6     3.19
#> # ... with 4,648 more rows, and 6 more variables: adjHigh <dbl>,
#> #   adjLow <dbl>, adjOpen <dbl>, adjVolume <int>, divCash <dbl>,
#> #   splitFactor <dbl>

# Nest monthly
aapl_monthly_nested <- aapl %>%
  mutate(ym = yearmonth(date)) %>%
[...]

PR #168, opened 2020-07-16
`sliding_*()` resamplers

### LINKED ISSUE #56: rolling_origin() should allow for an expanding window + sliding window in the same call
This Uber post talks about how useful it can be to expand a window up until you have enough data, and then proceed to use a sliding window. I think that could be useful and I don't think rolling_origin() does this at the moment. https://eng.uber.com/forecasting-introduction/

DavisVaughan:
This PR adds three new resampling functions, based on {slider}.

- `sliding_window()` is the most like `rolling_origin()`. It looks back a set number of rows to construct the analysis set, and looks forward to construct the assessment set. It doesn't deal with any time index whatsoever.

- `sliding_index()` is like `sliding_window()`, but it uses an `index` column to construct the resampling indices. For example, if the current row was on `2020-01-01`, the user could set `lookback = 20` to construct an analysis set that looks backwards "20 days" rather than 20 rows. This is useful if the `index` is irregular (has gaps).

- `sliding_period()` is my personal favorite. It takes the `index` column and a `period`, such as `"month"` or `"year"` and first constructs groups based on that period. Then it constructs the resampling indices based on those groups. Essentially this lets you create monthly / yearly resampling slices. It is a much better solution to https://github.com/tidymodels/rsample/issues/42

They are all in `R/slide.R`, which I think github won't expand by default since it is a little too long.

These generally replace `rolling_origin()`, but I don't think we need any immediate plans to deprecate it.

Closes #56 through the `complete` argument

---

An example of `sliding_period()` 

``` r
library(rsample)
library(modeldata)
data("Chicago")

# Break the resamples up by month:
# - Here we'll use an expanding monthly window by setting `lookback = Inf`
# - Each assessment set will contain two months of data.
# - To ensure that we have enough data to fit our models, we'll `skip` the
#   first 4 expanding windows.
# - Finally, to thin out the results, we'll `step` forward by 2 months between
#   each resample.
sliding_period(
  Chicago,
  date,
  "month",
  lookback = Inf,
  assess_stop = 2,
  skip = 4,
  step = 2
)
#> # Sliding period resampling 
#> # A tibble: 91 x 2
#>    splits           id     
#>    <list>           <chr>  
#>  1 <split [130/61]> Slice01
#>  2 <split [191/61]> Slice02
#>  3 <split [252/61]> Slice03
#>  4 <split [313/62]> Slice04
#>  5 <split [375/59]> Slice05
#>  6 <split [434/61]> Slice06
#>  7 <split [495/61]> Slice07
#>  8 <split [556/61]> Slice08
#>  9 <split [617/61]> Slice09
#> 10 <split [678/62]> Slice10
#> # … with 81 more rows
```

<sup>Created on 2020-07-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


DavisVaughan:
@mdancho84 I have created 3 new time based resampling functions based on slider. These should replace `rolling_origin()` (although we won't remove it).

I thought through the `lag` argument of `rolling_origin()`, but have come to the conclusion that it should not be included in these new functions. I am very worried about the fact that it results in data leakage between the training and testing set. Dealing with that leakage correctly is not very straightforward.

My proposed solution is that you should lag the original data before resampling and before passing it into the recipe. You would then not need a `step_lag()` in the recipe. This should not cause any problems, and results in no data leakage between training and test set.

More importantly, I believe that this matches how new data will appear in production when you are ready to call `predict()` with a fitted model. When I get 1 new row containing today's predictors to predict an outcome with, I will have to go ahead and include yesterday's lagged variables as columns in that row _before_ passing it on to `predict()` (which, for a workflow, would call `bake()` for preprocessing and do the prediction method specific to the model used). This implies that the lagged columns are not variables that result from recipe preprocessing, they are just raw predictors.

I have further explored these ideas in this gist:
https://gist.github.com/DavisVaughan/433dbdceb439c9be30ddcc78d836450d

mdancho84:
I agree 100%. 

For time series - The way I teach lags is to include this _before_ any recipe preprocessing. This way, when you generate new data (future data to predict on), you can easily add lagged variables to the future data. This is much better than trying to handle inside of a recipe. 

The lag argument approach was an early attempt at getting step_lag() to work. I couldn't get it to work because of the issues you bring up. 

So we've settled on the same conclusion - include lags outside of the recipe.

DavisVaughan:
I am very happy to hear that!

PR #170, opened 2020-07-28
Fix typo

hlynurhallgrims:
"An tibble" becomes "A tibble"


juliasilge:
Thank you so much @hlynurhallgrims! 🙌

Issue #177, opened 2020-09-03
alpha argument in int_bca does not change .upper and .lower limits


Dpananos:
The alpha parameter in the `int_bca` function does not return different interval estimates. See below.

### Minimal, reproducible example:

``` r

library(tidyverse)
#> Warning: package 'ggplot2' was built under R version 4.0.2
#> Warning: package 'tibble' was built under R version 4.0.2
#> Warning: package 'tidyr' was built under R version 4.0.2
#> Warning: package 'dplyr' was built under R version 4.0.2
library(rsample)
#> Warning: package 'rsample' was built under R version 4.0.2
library(broom)
#> Warning: package 'broom' was built under R version 4.0.2

d <- tibble(x = rnorm(25)) %>% 
  mutate(y = 2*x + 1,
         noise = rnorm(25),
         Y = y + noise)


fit_model<-function(splits, ...){
  lm(Y~x, data = analysis(splits)) %>% 
    tidy()
}

set.seed(0)
bt<-bootstraps(d, 2000, apparent = T) %>% 
    mutate(models = map(splits, fit_model)) 


int_bca(bt, models, .fn = fit_model, alpha = 0.5)
#> # A tibble: 2 x 6
#>   term        .lower .estimate .upper .alpha .method
#>   <chr>        <dbl>     <dbl>  <dbl>  <dbl> <chr>  
#> 1 (Intercept)  0.617     0.981   1.33    0.5 BCa    
#> 2 x            1.60      1.88    2.20    0.5 BCa
int_bca(bt, models, .fn = fit_model, alpha = 0.05)
#> # A tibble: 2 x 6
#>   term        .lower .estimate .upper .alpha .method
#>   <chr>        <dbl>     <dbl>  <dbl>  <dbl> <chr>  
#> 1 (Intercept)  0.617     0.981   1.33   0.05 BCa    
#> 2 x            1.60      1.88    2.20   0.05 BCa
```

<sup>Created on 2020-09-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### Session Info:

```r
─ Session info ──────────────────────────────────────────────────────────────────────────────────────────────────
 setting  value                       
 version  R version 4.0.0 (2020-04-24)
 os       macOS Catalina 10.15.4      
 system   x86_64, darwin17.0          
 ui       RStudio                     
 language (EN)                        
 collate  en_CA.UTF-8                 
 ctype    en_CA.UTF-8                 
[...]

PR #179, opened 2020-09-07
Bca bug calc

Dpananos:
In #177, @juliasilge mentions the bug may be because `alpha` is not passed to `get_p0`.  This PR fixes that and adds additional unit tests to ensure `int_bca` returns appropriate confidence limits when the confidence level changes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 * Three new time based resampling functions have been added: `sliding_window()`,
   `sliding_index()`, and `sliding_period()`, which have more flexibility than
   the pre-existing `rolling_origin()`.
+  
+* Correct passing `alpha` parameter for `int_bca()` (#179).
 
 # rsample 0.0.7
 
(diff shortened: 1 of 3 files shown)


Dpananos:
Hmm, not sure why tests are failing.  Doesn't look like its a problem with the code however.

Dpananos:
@juliasilge Is the change in tolerance necessary?  The bug didn't effect the estimates per se, it just didn't return the appropriate confidence limit intervals when an alpha other than 0.05 was passed.

juliasilge:
The tolerances are used to check both the estimates and the confidence intervals, like this:

```r
  expect_equal(ttest_lower_conf$conf.low,
               single_bca_res_lower_conf$.lower,
               tolerance = 0.001)
```

With the old tolerance, these tests of the high/low confidence values all pass even when `alpha` does not change.

Dpananos:
Great, thanks again!

juliasilge:
Thank you so much for the PR! 🚀

PR #182, opened 2020-09-11
Help documentation updates

irkaal:
Fixes typo in doc

### DIFF
--- a/R/slide.R
+++ b/R/slide.R
@@ -117,7 +117,7 @@
 #'   _in the time zone of the index_.
 #'
 #'   This is generally used to define the anchor time to count from,
-#'   which is relevant when the every value is `> 1`.
+#'   which is relevant when every value is `> 1`.
 #'
 #' @seealso
 #' [rolling_origin()]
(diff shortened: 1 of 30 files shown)


R/slide.R
```diff
@@ -117,7 +117,7 @@
 #'   _in the time zone of the index_.
 #'
 #'   This is generally used to define the anchor time to count from,
-#'   which is relevant when the every value is `> 1`.
+#'   which is relevant when every value is `> 1`.
```
DavisVaughan, line 120: This one was actually almost correct before, but it should have been

```
when the `every` value
```

`every` is a function argument, which is what i was talking about here. Can you please update this to what I showed above?

DavisVaughan (changes requested):
This looks great! We really appreciate these kinds of PRs. I have one comment below, but otherwise this is awesome, thanks!

R/slide.R
irkaal: Yes, of course. Sorry about that!

DavisVaughan:
Thank you!

Issue #183, opened 2020-09-14
Feature request: spatial cross-validation


MxNl:
After a rough research on wether it is possible to do cross-validation with spatial data with the `rsample` package I couldn't find any information. Thats why I would like to suggest a function for doing spatial cross-validation similar to `time_series_cv()` or add arguments for coordinate columns in `vfold_cv()`. Thanks for all your work!!!

juliasilge:
A few notes:

- mlr implements spatial resampling, [as described here](https://bookdown.org/robinlovelace/geocompr/spatial-cv.html#spatial-cv-with-mlr)
- The [sperrorest](https://cran.r-project.org/package=sperrorest) package has support for spatial resampling

DavisVaughan:
Also:

- https://github.com/rvalavi/blockCV
- https://github.com/mlr-org/mlr3spatiotempcv

topepo:
See also #32 and #64

juliasilge:
We are working on spatial resampling in our new package [spatialsample](https://github.com/tidymodels/spatialsample).

PR #185, opened 2020-09-22
Clustering approach to spatial CV

juliasilge:
Addresses #183.

This PR implements what I believe is the simplest and most straightforward approach that folks use for spatial cross validation. This is what the [sperrorest](https://github.com/giscience-fsu/sperrorest) package implemented (now no longer being actively developed) and what the [mlr group calls "spatial CV"](https://github.com/mlr-org/mlr3spatiotempcv#resampling-methods). This uses k-means clustering on the spatial variables to identify "chunks"; you can read about this [in this paper](https://doi.org/10.1109/IGARSS.2012.6352393); it is [what is shown in the plot here as well](https://bookdown.org/robinlovelace/geocompr/spatial-cv.html#intro-cv).

``` r
library(tidyverse)
library(rsample)
data("ames", package = "modeldata")

spatial_folds <- spatial_cv(ames, v = 5, repeats = 2, 
                            coords = c("Latitude", "Longitude"))


plot_splits <- function(split) {
  analysis(split) %>%
    mutate(analysis = "Analysis") %>%
    bind_rows(assessment(split) %>%
                mutate(analysis = "Assessment")) %>%
    ggplot(aes(Longitude, Latitude, color = analysis)) + 
    geom_point(alpha = 0.5) +
    labs(color = NULL)
}

plot_splits(spatial_folds$splits[[1]])
```

``` r
plot_splits(spatial_folds$splits[[2]])
```

``` r
plot_splits(spatial_folds$splits[[3]])
```

<sup>Created on 2020-09-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

If we wanted to do more complex spatial resampling like blocking, I think that will need to go in a separate package because of dependency issues like sf. This PR will be fine to keep in rsample itself.


R/spatial.R
```diff
@@ -0,0 +1,102 @@
...
+#' International Geoscience and Remote Sensing Symposium, Munich, 2012,
+#' pp. 5372-5375, doi: 10.1109/IGARSS.2012.6352393.
+#'
+#' @examples
+#' data(ames, package = "modeldata")
+#' spatial_cv(ames, v = 5, coords = c("Latitude", "Longitude"))
+#' spatial_cv(ames, v = 5, repeats = 2, coords = c("Latitude", "Longitude"))
+#'
+#' @export
+spatial_cv <- function(data, v = 10, repeats = 1, coords = NULL, ...) {
```
DavisVaughan, line 47: `coords` are required? How about:

```
spatial_cv <- function(data, coords, v = 10, repeats = 1, ...) {
```

Where `coords` can be specified using tidyselect, and you can capture that with:

```
tidyselect::eval_select(rlang::enquo(coords), data = data)
```

So then the API would be:

```
spatial_cv(ames, c(Latitude, Longitude), v = 5, repeats = 2)
```

Personally in cases like this where you have multiple required args and then a few optional ones, I'm partial to placing the `...` between them to force the user to name the optional args. So you could even do:

```
spatial_cv <- function(data, coords, ..., v = 10, repeats = 1) {
```

and then use `ellipsis::check_dots_empty()` to check for misspecifications. This also allows us to easily add new optional arguments to `spatial_cv()` later on without fear of breaking existing code (since the optional args all have to be specified by name, not position)

topepo: I think that `...` should go to `kmeans()` since there are some important options there so I'm ok with `coords`.

R/spatial.R
```diff
@@ -0,0 +1,102 @@
...
+
+  if(missing(coords) | !all(coords %in% names(data))) {
+    rlang::abort("`coords` must be variables in `data`.")
+  }
+
+  if (repeats == 1) {
+    split_objs <- spatial_splits(data = data, v = v, coords = coords)
+  } else {
+    for (i in 1:repeats) {
```
topepo, line 56: What would repeats do? Since it's not really random (since the allocations are set by k-means), I think that you would get the same folds on repeat. 

Here you go: 

image

R/spatial.R
```diff
@@ -0,0 +1,102 @@
+#' Spatial or Cluster Cross-Validation
```
topepo, line 1: If we are going to add more spatial methods to a side-package (since it would require `sf`), this would probably go there instead of `rsample`. 

Also, if we are doing that, a less generic name would be better. `spatial_clustering_cv()`?

topepo (changes requested):

R/spatial.R
juliasilge: Without a seed they are not _exactly_ the same and sort of show the variability of this approach. What do you all think? `repeats` or no `repeats`?

R/spatial.R
juliasilge: FWIW in other packages, posts, books, etc this approach seems to be called "spatial CV" while other approaches are called things like "block CV" or "buffer CV", etc.

R/spatial.R
juliasilge: We decided to remove `repeats` since it is so likely that repeats will be so similar.

MxNl:
Thank you for implementing this in rsample. I also think it would be really useful to add the possibility to additionally stratify the spatial cross-validation. Since geodata often has categorical features, there is a need to use stratification when resampling. Otherwise, samples of locally distinct categorical classes might end up either exclusively in the training or validation split. Ideally, stratification would be possible for several categorical characteristics. What do you think?

topepo:
@MxNl That might be tough; we'd have to encode that information into the k-means process so that it is part of the decision. We could add other columns to the data but that might severely dilute the process and would be especially problematic if the other characteristics themselves have spatial effects (very likely). 

We're currently trying to decide if this PR goes into `rsample` or another package. Once that get's figured out, I'll ping back here and we can add another issue for this feature.

tiernanmartin:
Any luck finding a home for this feature?

I see that @topepo [reached out](https://github.com/rvalavi/blockCV/issues/15) to the owner of the {[blockCV](https://github.com/rvalavi/blockCV)} package, but it seems like that developer hasn't been very active on that repo or Github in general for the past three months.

juliasilge:
@tiernanmartin Our current plan (since the blockCV folks don't seen interested in working together right now) is to create an rsample-adjacent package with spatial resampling methods. It will include the simpler k-means CV that is in this branch, and also the more complex CV methods that require sf dependencies (we don't want rsample to take the sf/gdal dependencies).

Until we get that going, you can use the current spatial CV by [installing from this branch](https://devtools.r-lib.org/reference/install_github.html):

```r
devtools::install_github("tidymodels/rsample@spatial-cv")

## then
## data(ames, package = "modeldata")
## rsample::spatial_clustering_cv(ames, v = 5, coords = c(Latitude, Longitude))
```

juliasilge:
I've got the new package [spatialsample](https://github.com/tidymodels/spatialsample) up and running, so we are closing this PR.

PR #193, opened 2020-10-09
Use automatic pkgdown

### LINKED ISSUE #191: Update pkgdown docs to be automatic
If possible. At least update them so sliding-window and friends are in the production docs

DavisVaughan:
Closes #191


DavisVaughan:
*sigh* won't be able to tell if it works right now because of the textshaping install issues

PR #199, opened 2020-10-19
Update contributing info

juliasilge:
This PR removes the vestigial issue template and adds the `CONTRIBUTING.md` like other core tidymodels packages.


juliasilge:
This is failing because of recipes being not available for R < 3.5, due to lava. Should we change testing on rsample for now, [like recipes](https://github.com/tidymodels/recipes/commit/133ea628b2eee83c8553ec61aed1749c87947cf6)?

PR #200, opened 2020-10-23
Added permutation resampling functions

### LINKED ISSUE #198: Feature request: Permutation resampling
Hi,

I propose the addition of a function for row-wise permutation-based resampling. Happy to have a discussion about whether it fits into the vision of `rsample`. Basically, I want to consider resurrecting #13. `recipes::step_shuffle()` is helpful but not quite sufficient for large scale permutation resampling. `infer::generate()` is closer to what is desired, but it doesn't exactly follow the `rsample` paradigm. I should add, if there is interest, I would be happy to contribute this feature via a PR. 

### Motivation
- I think this type of resampling function fits into the natural ecosystem with `rsample`
- I am a proponent of using permutation-based resampling for generating null distributions for statistics (e.g. p-values), such as when the null distribution is unknown or if distributional assumptions for parametric tests may be violated
- _Counterpoint_: Unlike the other resampling methods in `rsample` which serve to generate samples that have the appearance of being new draws from the same underlying data-generating mechanism (or the alternative dist.), permutation resampling serves to generate samples under the null. Despite this difference, I still think it possibly fits in here.
- _Issue_: Permutation sampling doesn't have defined splits, so what would `training()`/`testing()` or `analysis()`/`assessment()` return? This is a bit of a sticking point, perhaps.

### Considerations
- Function could be named `permutations()` to have the same feel as `bootstraps()`, but open for suggestions
- Could function in one of two ways:
  1. Return a fixed number of permutations, akin to `bootstraps()`
  2. Return all possible permutations, as is common for permutation tests (could be an enormous amount of permutations for large data/multiple columns). In most cases one would only need to permute the response variable which simplifies things a great deal...
- I think the function should also include a `strata` argument to perform stratified permutation; probably should also have all the same arguments as bootstraps
- I think the function should maybe contain a `cols` argument specifying which columns should be permuted, while all other columns remain in their natural order. `cols` could accept `tidyselect` functions to select columns to permute.
- Could add some helper functions to:
  + Summarize permutation-based statistics with mean, SE, and confidence intervals (similar to `int_pctl()`)
  + Plot the permutation distribution of a chosen statistic and show where the apparent value falls in relation

### Proposed API
For `bootstrap`-like functionality...
```r
permutations(data, times = 25, strata = NULL, cols = everything(), breaks = 4, apparent = FALSE, ...)
```
For permutation-test like functionality...
```r
permutations(data, strata = NULL, cols = everything(), breaks = 4, apparent = FALSE, ...)
```

mattwarkentin:
Closes #198


mattwarkentin:
GitHub Action failures are the same as #199. Unrelated to my PR.

R/permutations.R
```diff
@@ -0,0 +1,116 @@
...
+  ...
+) {
+  permute <- rlang::enquo(permute)
+  if (is.null(permute)) {
+    stop("You must specify at least one column to permute!")
+  }
+  col_id <- tidyselect::eval_select(permute, data)
+
+  if (identical(length(col_id), 0L)) {
+    stop("You must specify at least one column to permute!")
```
topepo, line 62: Please use `rlang::abort()`

R/rsplit.R
```diff
@@ -90,6 +90,15 @@ as.data.frame.rsplit <-
     warning( "`optional` is kept for consistency with the ",
              "underlying class but TRUE values will be ",
              "ignored.", call. = FALSE)
+  if (!is.null(x$col_id)) {
```
topepo, line 93: Can you change the message to use the class of `x` instead of hard-coding it?

R/permutations.R
```diff
@@ -0,0 +1,116 @@
...
+#' permutations(mtcars, mpg, times = 2)
+#' permutations(mtcars, mpg, times = 2, apparent = TRUE)
+#'
+#' library(purrr)
+#' resample1 <- permutations(mtcars, starts_with("c"), times = 1)
+#' resample1$splits[[1]] %>% analysis()
+#'
+#' resample2 <- permutations(mtcars, hp, times = 10, apparent = TRUE)
+#' map_dbl(resample2$splits, function(x) {
+#'   t.test(hp ~ vs, data = analysis(x))$p.value
```
topepo, line 44: I would return the estimate instead of the p-value.

topepo: Actually, the `statistic` field.

R/permutations.R
```diff
@@ -0,0 +1,116 @@
...
+#' @description
+#' A permutation sample is the same size as the original data set and is made
+#'   by permuting/shuffling one or more columns. This results in analysis
+#'   samples where some columns are in their original order and some columns
+#'   are permuted to a random order. Unlike other sampling functions in
+#'   `rsample`, there is no assessment set and calling `assessment()` on a
+#'   permutation split will throw an error.
+#'
+#' @param data A data frame.
+#' @param permute <`tidy-select`> One or more unquoted
```
topepo, line 12: I'm not a fan of this documentation style. Can you remove "<`tidy-select`>"?

topepo (changes requested):
Can you update the NEWS and `_pkgdown.yml` files too?

R/permutations.R
mattwarkentin: Sure, no problem. I was trying to be consistent with how `dplyr` and `tidyr` often describe accepting `tidyselect` selectors. But happy to change.

mattwarkentin:
Completed all suggested revision, @topepo. I also added the `col_id` as an attribute to the `tibble` as a whole. I think it makes it easy for users to double-check which columns which columns were shuffled. I also think that if `rsample` print messages ever get a makeover with `cli`, then this attribute would be useful for printing more informative messages.

Please let me know if you would like this reverted.

topepo:
That's a good idea. Can you add some tests? We should make sure that: 

 * filtering rows reverts the class to tibble and
 * column binding a data set keeps the default classes.

mattwarkentin:
Hey Max,

I added tests for class inheritance for filtering and slicing rows, and also different ways of column binding involving a permutation `rset` object. I also went ahead and made the update to the `print` method as shown below. More so than with other resample methods, I think its really important to know details about how the permutation was performed (i.e which columns were shuffled), so I thought making it more obvious with the `print` method was important. Let me know what you think! 

``` r
library(rsample)
permutations(mtcars, 1:3)
#> # Permutation sampling 
#> # Permuted columns: [mpg, cyl, disp] 
#> # A tibble: 25 x 2
#>    splits         id            
#>    <list>         <chr>         
#>  1 <split [32/0]> Permutations01
#>  2 <split [32/0]> Permutations02
#>  3 <split [32/0]> Permutations03
#>  4 <split [32/0]> Permutations04
#>  5 <split [32/0]> Permutations05
#>  6 <split [32/0]> Permutations06
#>  7 <split [32/0]> Permutations07
#>  8 <split [32/0]> Permutations08
#>  9 <split [32/0]> Permutations09
#> 10 <split [32/0]> Permutations10
#> # … with 15 more rows
```

topepo:
I think that it's a good idea.

topepo:
I think that an update to the `_pkgdown.yml` file is all that is left.

mattwarkentin:
Already done in 2423137 [(see here)](https://github.com/tidymodels/rsample/commit/24231371cbdaa8558d8ee5dd7d88c8530bee7a31#diff-f5f791eb2285a738ec729c44088e49a60ddf355eb479fb1fb5df7fcea4fea2e8)

PR #201, opened 2020-10-30
Fingerprinting rset objects

topepo:
@simonpcouch You might consider using this in `stacks`.


R/misc.R
```diff
@@ -57,3 +57,55 @@ split_unnamed <- function(x, f) {
...
+  id_vars <- grep("^id", names(x), value = TRUE)
+  if (length(id_vars) == 0) {
+    rlang::abort("No ID columns were found.")
+  }
+  if (!any(names(x) == "splits")) {
+    rlang::abort("The 'split' column was not found.")
+  }
+
+  x <-
+    dplyr::select(x, splits, dplyr::matches("^id")) %>%
```
DavisVaughan, line 100: I would probably use:

```r
names <- names(x)
detect_id <- col_starts_with_id(names)
names_id <- names[detect_id]
```

then:

```r
dplyr::select(x, splits, all_of(names_id))
```

to avoid rematching and to use the more complex regex in `col_starts_with_id()`

R/misc.R
```diff
@@ -57,3 +57,55 @@ split_unnamed <- function(x, f) {
...
+    dplyr::distinct() %>%
```
DavisVaughan, line 101: ~Is there a specific use case you had in mind where this would be useful? On first thought I figured I'd want duplicates to give a different hash.~

Update: On looking at the tests, it seems this is for `tune_bayes()`. I guess that makes sense

---

I do think arranging probably makes sense though

R/misc.R
```diff
@@ -57,3 +57,55 @@ split_unnamed <- function(x, f) {
...
+    dplyr::arrange(!!!id_vars) %>%
+    tibble::as_tibble()
+  attr_x <- names(attributes(x))
+  attr_rm <- attr_x[!(attr_x %in% c("row.names", "names", "class"))]
+  for (i in attr_rm) {
+    attr(x, i) <- NULL
+  }
```
DavisVaughan, line 108: Slightly clearer, 

```r
attrib <- attributes(x)
attrib <- attrib[names(attrib) %in% c("row.names", "names", "class")]
attributes(x) <- attrib
```

DavisVaughan (approved):

PR #205, opened 2020-11-30
Add id to initial_time_split()

### LINKED ISSUE #175: last_fit() error because initial_time_split() has no id property
I had to add the 'id' property to initial_split object myself to prevent error in last_fit() step.

``` r
library(tidymodels)

mtcars_initial_split <- mtcars %>% initial_time_split(prop = 0.5)
# mtcars_initial_split$id <- tibble(id = "Resample1") # <----- error in tune:::append_metrics(). uncomment this line to prevent the error.

mtcars_model <- linear_reg(mode = "regression") %>% set_engine("lm")
mtcars_last_fit <- last_fit(mtcars_model, mpg ~ ., split = mtcars_initial_split)
#> x : internal: Error in data.frame(..., check.names = FALSE): arguments imply...
#> Warning: All models failed in [fit_resamples()]. See the `.notes` column.
```

<sup>Created on 2020-08-26 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

juliasilge:
Closes #175 

``` r
library(rsample)

split1 <- initial_split(mtcars)
split1$id
#> # A tibble: 1 x 1
#>   id       
#>   <chr>    
#> 1 Resample1

split2 <- initial_time_split(mtcars)
split2$id
#> # A tibble: 1 x 1
#>   id       
#>   <chr>    
#> 1 Resample1
```

<sup>Created on 2020-11-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

I don't know if we want to make this change as shown here with `manual_rset()`, or just by adding the `id` in by assignment. Thoughts?

(By the way, the problem with `last_fit()` as shown in the linked issue is no longer a problem as of the new tune release, even without the `id`.)

### DIFF
--- a/R/initial_split.R
+++ b/R/initial_split.R
@@ -80,7 +80,13 @@ initial_time_split <- function(data, prop = 3/4, lag = 0, ...) {
     stop("`lag` must be less than or equal to the number of training observations.", call. = FALSE)
   }
 
-  rsplit(data, 1:n_train, (n_train + 1 - lag):nrow(data))
+  res <- manual_rset(
+    splits = list(rsplit(data, 1:n_train, (n_train + 1 - lag):nrow(data))),
+    ids = "Resample1"
+  )
+
+  res$splits[[1]]
+
 }
 
 #' @rdname initial_split
--- a/R/initial_split.R
+++ b/R/initial_split.R
@@ -80,12 +80,12 @@ initial_time_split <- function(data, prop = 3/4, lag = 0, ...) {
     stop("`lag` must be less than or equal to the number of training observations.", call. = FALSE)
   }
 
-  res <- manual_rset(
-    splits = list(rsplit(data, 1:n_train, (n_train + 1 - lag):nrow(data))),
-    ids = "Resample1"
-  )
+  split  <- rsplit(data, 1:n_train, (n_train + 1 - lag):nrow(data))
+  splits <- list(split)
+  ids    <- "Resample1"
+  rset   <- new_rset(splits, ids)
 
-  res$splits[[1]]
+  rset$splits[[1]]
 
 }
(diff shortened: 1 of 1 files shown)


juliasilge:
Chatting with Max today, we both said we lean toward using the "constructor"-ish `manual_rset()`, although it is sort of A LOT.

R/initial_split.R
```diff
@@ -80,7 +80,13 @@ initial_time_split <- function(data, prop = 3/4, lag = 0, ...) {
     stop("`lag` must be less than or equal to the number of training observations.", call. = FALSE)
   }
 
-  rsplit(data, 1:n_train, (n_train + 1 - lag):nrow(data))
+  res <- manual_rset(
```
DavisVaughan, line 83: The reason this even works is that `manual_rset()` eventually calls `new_rset()`, and `new_rset()` is what adds the id bits to the splits. So really all you need here is `new_rset()`. You can probably do something like:

```r
split <- rsplit(data, 1:n_train, (n_train + 1 - lag):nrow(data))
splits <- list(split)
ids <- "Resample1"
rset <- new_rset(splits, ids)
rset$splits[[1]]
```

That would be fine for now, but I'm pretty convinced that we shouldn't be adding the id in `new_rset()`, it should be done in `rsplit()`. Then it would just be:

```r
rsplit(data, 1:n_train, (n_train + 1 - lag):nrow(data), id = "Resample1")
```

Something to think about at another time I guess

DavisVaughan (changes requested):

R/initial_split.R
juliasilge: Thank you! 👍

PR #206, opened 2021-01-13
A simple function to compute bootstrap intervals for commonly used regression models

topepo:
Based on some thoughts that @mine-cetinkaya-rundel gave us from her teaching. 

Works for `lm()`, `glm()`, `survreg()`,  and `coxph()` models

``` r
library(rsample)

set.seed(1)
reg_intervals(mpg ~ ., data = mtcars)
#> # A tibble: 10 x 6
#>    term   .lower .estimate .upper .alpha .method  
#>    <chr>   <dbl>     <dbl>  <dbl>  <dbl> <chr>    
#>  1 am    -3.92      2.15   9.68     0.05 student-t
#>  2 carb  -2.35     -0.495  2.86     0.05 student-t
#>  3 cyl   -3.47      0.162  2.98     0.05 student-t
#>  4 disp  -0.0345    0.0150 0.0649   0.05 student-t
#>  5 drat  -5.15      1.87   4.22     0.05 student-t
#>  6 gear  -4.85      1.47   4.22     0.05 student-t
#>  7 hp    -0.104    -0.0130 0.0316   0.05 student-t
#>  8 qsec  -1.95      1.33   2.70     0.05 student-t
#>  9 vs    -4.42     -0.584  6.66     0.05 student-t
#> 10 wt    -9.41     -3.98   2.15     0.05 student-t
```

<sup>Created on 2021-01-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


R/reg_intervals.R
```diff
@@ -0,0 +1,105 @@
...
+#' @param data A data frame.
+#' @param model_fn The model to fit. Allowable values are "lm", "glm",
+#'  "survreg", and "coxph". The latter two require that the `survival` package
+#'  be installed.
+#' @param type The type of bootstrap confidence interval. Values of "student-t" and
+#' "percentile" are allowed.
+#' @param times A single integer for the number of bootstrap samples. If left
+#' NULL, 1,001 are used for t-intervals and 2,001 for percentile intervals.
+#' @param alpha Level of significance.
+#' @param filter A logical statement to filter the model terms (or NULL).
```
DavisVaughan, line 13:
```suggestion
#' @param filter A logical expression used to remove rows from the final result, or `NULL` to keep all rows.
```

You might want to tweak it a little more, but generally I wanted to get at the fact that this removes _rows_ from the final result using a `filter()`

R/reg_intervals.R
```diff
@@ -0,0 +1,105 @@
...
+#' @param keep_reps Should the individual parameter estimates for each bootstrap
+#' sample be retained?
+#' @param ... Options to pass to the model function (such as `family` for `glm`).
```
DavisVaughan, line 16:
```suggestion
#' @param ... Options to pass to the model function (such as `family` for `glm()`).
```

R/reg_intervals.R
```diff
@@ -0,0 +1,105 @@
...
+#' reg_intervals(mpg ~ I(1/sqrt(disp)), data = mtcars)
+#'
+#' set.seed(1)
+#' reg_intervals(mpg ~ I(1/sqrt(disp)), data = mtcars, keep_reps = TRUE)
+#' }
+reg_intervals <-
+  function(formula, data, model_fn = "lm", type = "student-t", times = NULL,
+           alpha = 0.05, filter = term != "(Intercept)", keep_reps = FALSE, ...) {
+    model_fn <- match.arg(model_fn, c("lm", "glm", "survreg", "coxph"))
+    type <- match.arg(type, c("student-t", "percentile"))
```
DavisVaughan, line 41: Also consider `rlang::arg_match()` for better error messages and no partial matching

R/reg_intervals.R
```diff
@@ -0,0 +1,105 @@
...
+      )
+    if (type == "student-t") {
+      res <- int_t(bt, models, alpha = alpha)
+    } else {
+      res <- int_pctl(bt, models, alpha = alpha)
+    }
+
+    if (keep_reps) {
+      bt <- bt[bt$id != "Apparent",]
+      reps <- purrr::map_dfr(bt$models, I)
```
DavisVaughan, line 87: What does this do?

topepo: It's a cheap way of unnesting the model estimates list column. I could use `map_dfr(bt$models, ~ x)`.

mattwarkentin:
Hey @topepo and @DavisVaughan,

Just poking my head in here to ask about whether or not this function could be more general and thus have more widespread application. It seems to me that following the same logic already contained in the current code, this function could be made to work for any model that has a formula method for model fitting and a tidy method, which is quite a large number of models.

I am thinking the function API could look more like:
```r
# Default args
reg_intervals(formula, data, model_fn = base::lm, tidier = broom::tidy, ...)
```

```r
# A simple lm() example would look the same
reg_interval(mpg ~ ., mtcars)

# A more "unconventional" model example would still work
reg_interval(mpg ~ ., mtcars, model_fn = mlogit::mlogit)

# Could even work for models without formal tidiers by supplying your own tidier
reg_interval(
  mpg ~ ., data = mtcars, 
  model_fn = mypkg::very_special_model, 
  tidier = function(x) {some code that returns a tibble of coefficients}
)
```

The suggested API could piggy-back on all of the existing `tidy()` methods which would be nice, while still remaining somewhat agnostic to how the data model results get tidy.

This type of usage may be beyond the scope of what this function is supposed to be, but I'm just thinking out loud. Perhaps it deserves to be its own function, which I'd be happy to work on, if desired.

topepo:
It could but I'm wary of supporting everything. 

For example, right now, we will need to add some kind of filter for models to check to see if they converged or not. That is very difficult to do with `survival` models (as an example) and might lead to some issues.

PR #208, opened 2021-01-20
Add new exported `complement_default()` function

juliasilge:
This PR adds a new `complement_default()` function, to be used by [spatialsample](https://github.com/tidymodels/spatialsample) in `complement` methods, and anything else in the future.

``` r
library(rsample)
df <- data.frame(x = c(1:10))
indices <- list(analysis = 1:6, assessment = 7:10)

split <- make_splits(indices, df)
complement_default(split)
#> [1]  7  8  9 10
```

<sup>Created on 2021-01-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

We decided that other small functions would be fine to copy (like `names0()` and such).


juliasilge:
Closed by #209

PR #216, opened 2021-02-04
New inheritance structure for `rsplit` objects and updated `complement()` methods

hfrick:
This is to address #213 as laid out in [this comment](https://github.com/tidymodels/rsample/issues/213#issuecomment-771637109).

PR #218, opened 2021-02-10
removed `at_end` arg

hfrick:
we were not using it and likely will not either in future (discussed in #216)


DavisVaughan (approved):
Thanks!

Issue #217, opened 2021-02-07
initial_split() with strata argument behavior


issactoast:
## The problem

I was expecting to have the same dimension of dataframe from 
Unexpected size of data frame after initial_split() with strata argument compare to sample_frac()
At least, the original nrow of airquality data is 153 so I guess `airquality_frac2 ` should have 15 or 16.

## Reproducible example

``` r
#sample_frac and sample_n
library(dplyr)
#> 
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

set.seed(2021)
dim(airquality)
#> [1] 153   6

#sample 10% from each month
airquality_frac <- airquality %>% 
    group_by(Month) %>%  
    sample_frac(size = 0.1, replace=FALSE)

library(rsample)

airquality_split <- initial_split(airquality, prop = 0.1,
              strata = "Month")

airquality_frac2 <- training(airquality_split)

dim(airquality_frac)
#> [1] 15  6
dim(airquality_frac2)
#> [1] 18  6
```

juliasilge:
This is because of the stratification and how we use `floor(n * prop)` in `strat_sample()`:
https://github.com/tidymodels/rsample/blob/54e3235a36627b736c4eb3b663eabe6c2061e5e5/R/mc.R#L114-L120

The quantity `airquality$Month` is not uniformly distributed in the sample, especially compared to the overall size, so when we do stratified sampling within each month, we get that result you see.

``` r
library(rsample)
table(airquality$Month)
#> 
#>  5  6  7  8  9 
#> 31 30 31 31 30

initial_split(airquality, prop = 0.1)
#> <Analysis/Assess/Total>
#> <16/137/153>
initial_split(airquality, prop = 0.9)
#> <Analysis/Assess/Total>
#> <138/15/153>

initial_split(airquality, prop = 0.1, strata = Month)
#> <Analysis/Assess/Total>
#> <18/135/153>
initial_split(airquality, prop = 0.9, strata = Month)
#> <Analysis/Assess/Total>
#> <140/13/153>
```

<sup>Created on 2021-02-08 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

We get those results like 13 and 18 from adding up the results of multiplying numbers like `31 * 0.9` and then using `floor()` _within_ the strata.

[...]

PR #219, opened 2021-02-16
PR related to issue #217.

issactoast:
Hi @juliasilge 

This PR related to issue #217 

I have been thinking about this since I found `strat_sample` sounds good to me, so I fork the package and let my data flow the functions. 

I found two parts is switched which make the result different than `sample_frac` in `dplyr`

I twicked `mc_splits` input from 1-prop to prop, so that the functions in mc.R find the index for prop. And I finally use `ind` for the analysis. 

These two chance makes the result from rsample and dplyr to be the same as follows;

``` r
#sample_frac and sample_n
library(dplyr)

set.seed(2021)
dim(airquality)
#> [1] 153   6

#sample 10% from each month
airquality_frac <- airquality %>% 
  group_by(Month) %>%  
  sample_frac(size = 0.1, replace=FALSE)

library(rsample)

airquality_split <- initial_split(airquality, prop = 0.1,
                                  strata = "Month")

airquality_frac2 <- training(airquality_split)

dim(airquality_frac)
#> [1] 15  6
dim(airquality_frac2)
#> [1] 15  6
```

Could you give me a feedback? I have feeling that this resolves the issue (user expectation about prop input) and increases the coherence between the tidymodels and the tidyverse packages. What do you think?

### DIFF
--- a/R/mc.R
+++ b/R/mc.R
@@ -61,7 +61,7 @@ mc_cv <- function(data, prop = 3/4, times = 25, strata = NULL, breaks = 4, ...) 
 
   split_objs <-
     mc_splits(data = data,
-              prop = prop, # prop for train set
+              prop = prop,
               times = times,
               strata = strata,
               breaks = breaks)
@@ -81,9 +81,9 @@ mc_cv <- function(data, prop = 3/4, times = 25, strata = NULL, breaks = 4, ...) 
            subclass = c("mc_cv", "rset"))
 }
 
-# Get the indices of the analysis set from the assessment set
+# Get the indices of the assessment set from the analysis set
 mc_complement <- function(ind, n) {
-  list(analysis = ind, # ind for analysis
+  list(analysis = ind,
        assessment = setdiff(1:n, ind))
 }
 
--- a/R/mc.R
+++ b/R/mc.R
@@ -61,7 +61,7 @@ mc_cv <- function(data, prop = 3/4, times = 25, strata = NULL, breaks = 4, ...) 
 
   split_objs <-
     mc_splits(data = data,
-              prop = 1 - prop,
+              prop = prop, # prop for train set
               times = times,
               strata = strata,
               breaks = breaks)
@@ -83,8 +83,8 @@ mc_cv <- function(data, prop = 3/4, times = 25, strata = NULL, breaks = 4, ...) 
 
 # Get the indices of the analysis set from the assessment set
 mc_complement <- function(ind, n) {
-  list(analysis = setdiff(1:n, ind),
-       assessment = ind)
+  list(analysis = ind, # ind for analysis
+       assessment = setdiff(1:n, ind))
 }
 
 
--- a/R/validation_split.R
+++ b/R/validation_split.R
@@ -35,7 +35,7 @@ validation_split <- function(data, prop = 3/4, strata = NULL, breaks = 4, ...) {
 
   split_objs <-
     mc_splits(data = data,
-              prop = prop, # prop for analysis
+              prop = prop,
               times = 1,
               strata = strata,
               breaks = breaks)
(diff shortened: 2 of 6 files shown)


issactoast:
I resolved the testthat package issue and the prop change in validation_split.R

R/mc.R
```diff
@@ -61,7 +61,7 @@ mc_cv <- function(data, prop = 3/4, times = 25, strata = NULL, breaks = 4, ...)
 
   split_objs <-
     mc_splits(data = data,
-              prop = 1 - prop,
+              prop = prop, # prop for train set
```
DavisVaughan, line 64: Could you please remove the comment? We can always look back at this PR and the associated issue if we are wondering why the code is written this way

R/mc.R
```diff
@@ -83,8 +83,8 @@ mc_cv <- function(data, prop = 3/4, times = 25, strata = NULL, breaks = 4, ...)
 
 # Get the indices of the analysis set from the assessment set
 mc_complement <- function(ind, n) {
-  list(analysis = setdiff(1:n, ind),
-       assessment = ind)
+  list(analysis = ind, # ind for analysis
```
DavisVaughan, line 86: Same, no comment needed

R/mc.R
```diff
@@ -83,8 +83,8 @@ mc_cv <- function(data, prop = 3/4, times = 25, strata = NULL, breaks = 4, ...)
 
 # Get the indices of the analysis set from the assessment set
```
DavisVaughan, line 84: Could you please update this comment to say:

```
# Get the indices of the assessment set from the analysis set
```

R/validation_split.R
```diff
@@ -35,7 +35,7 @@ validation_split <- function(data, prop = 3/4, strata = NULL, breaks = 4, ...) {
 
   split_objs <-
     mc_splits(data = data,
-              prop = 1 - prop,
+              prop = prop, # prop for analysis
```
DavisVaughan, line 38: Same, no comment required

tests/testthat/test_validation.R
```diff
@@ -49,8 +49,8 @@ test_that('strata', {
   rs3 <- validation_split(warpbreaks, strata = "tension")
   sizes3 <- dim_rset(rs3)
 
-  expect_true(all(sizes3$analysis == 42))
-  expect_true(all(sizes3$assessment == 12))
+  expect_true(all(sizes3$analysis == 39))
+  expect_true(all(sizes3$assessment == 15))
```
DavisVaughan, line 53: Could you please also the following test to `test_initial.R`:

```r
test_that("`prop` computes the proportion for analysis (#217)", {
  set.seed(11)
  
  props <- c(.1, .9)
  
  for (prop in props) {
    # Not stratified
    split <- initial_split(airquality, prop = prop)
    actual <- nrow(analysis(split))
    expect <- as.integer(floor(nrow(airquality) * prop))
    
    expect_identical(actual, expect)
    
    # Stratified
    split <- initial_split(airquality, prop = prop, strata = Month)
    actual <- nrow(analysis(split))
    expect <- as.integer(sum(floor(table(airquality$Month) * prop)) )
    
    expect_identical(actual, expect)
  }
})
```

This should ensure that this can't happen again. Adding a test like this is also a more robust alternative for the comments you had added directly in the code (which I requested you remove)

DavisVaughan (changes requested):
Thanks @issactoast! This looks like the right thing to do (as noted in https://github.com/tidymodels/rsample/issues/217#issuecomment-782151875). I have a few comments. Once you make these changes, I'll merge.

Notes to self:
- Rebase on master before merging
- Add a NEWS bullet (wait until after rebasing to keep this news entry separate from the 0.0.9 ones)

issactoast:
I have 
- removed and updated the comments on mc.R and  validation_split.R.
- added testthat code to the test_initial.R

Wow, this PR is my first contribution to the R community! It feels awesome! Thank you @DavisVaughan and @juliasilge! :)

DavisVaughan:
Perfect, I'll take it from here!

DavisVaughan:
Thanks @issactoast! That was a very useful first contribution!

issactoast:
@DavisVaughan Wonderful! Thank you for the concrete comment! 👍

PR #234, opened 2021-04-27
Update slide.R

mattwarkentin:
I could be wrong, but I think you mean "coarse" not "granular". The original wording gave the opposite impression of what is actually happening.

### DIFF
--- a/R/0_imports.R
+++ b/R/0_imports.R
@@ -16,8 +16,9 @@
 # ------------------------------------------------------------------------------
 
 utils::globalVariables(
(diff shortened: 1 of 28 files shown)


DavisVaughan:
Could you please say `less granular`?

Could you also please run `devtools::document()`?

mattwarkentin:
Whoops, I meant to use `/document`. I made this minor edit right on GitHub so I couldn't document the normal way, but I think once the `/document`-triggered GitHub Action finished it will push the changes here.

mattwarkentin:
Okay, this PR turned into a nightmare, let me take a redo on this one. Brb.

PR #242, opened 2021-05-19
Correct issue link

rkb965:
Correct link to issues that currently points to `parsnip` rather than `rsample`


juliasilge (approved):
Thank you so much @rkb965! 🙌

PR #245, opened 2021-06-07
Clarify discussion of `strata`

juliasilge:
This PR clarifies how `strata` works, especially with regards to numeric stratification variables. It uses templating to minimize repetition.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -28,13 +28,13 @@ jobs:
         config:
           - {os: macOS-latest,   r: 'release'}
           - {os: windows-latest, r: 'release'}
-          - {os: windows-latest, r: '3.6'}
+          - {os: windows-latest, r: 'oldrel'}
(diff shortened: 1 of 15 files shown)


hfrick (approved):
the reduction of repetition is rather nice :heart_eyes:

samuelmacedo83:
Hi @juliasilge, 
I was reading the documentation about `make_strata` and `vfold_cv` and I'm a lit bit confuse. 
In `make_strata`, the documentation said that numeric variables are divided by [percentile](https://github.com/tidymodels/rsample/blob/62902cb6d9fee3a3b20b24103e5a7b27a6ba5b51/R/make_strata.R#L16), but in `vfold_cv` the numeric variables are divided by [quartile](https://github.com/tidymodels/rsample/blob/62902cb6d9fee3a3b20b24103e5a7b27a6ba5b51/R/vfold.R#L19).

So, even if I use, for example, a 10 fold cv, the numeric variable will be divided by quartiles?

juliasilge:
The function `make_strata()` uses the `stats::quantile()` function with a default of `breaks = 4`, which means quartiles:
https://github.com/tidymodels/rsample/blob/62902cb6d9fee3a3b20b24103e5a7b27a6ba5b51/R/make_strata.R#L119

You can create different bins (e.g. use different percentiles of the data) by changing `breaks`. Maybe we can clarify the `make_strata()` documentation a bit more.

samuelmacedo83:
I think I understood. So the `breaks` parameter on `vfold_cv` is related to how to stratify the variable defined on  `strata` parameter. 
It makes sense :)
thank you @juliasilge

PR #270, opened 2021-11-12
Support variables in args for `nested_cv()` call objects

### LINKED ISSUE #81: How to pass a variable to group_vfold_cv's partition number argument
Hi,
I like to make a few `nested_cv`'s based on the same partition configuration as follows:

```
outer_cv <- 5
inner_cv <- 4
sampling1 <- nested_cv(all_dataset,
                      outside = group_vfold_cv(v = outer_cv, group = "Rep"),
                      inside = group_vfold_cv(v = inner_cv, group = "Rep"))
sampling2 <- ...
...
```

However, I am getting 

> object 'outer_cv' not found

 error, which is out of the scope for the `group_vfold_cv` function.
Do you have any recommendations? Does tidy evaluation help in this case?

juliasilge:
Closes #81 

I spent some time today working on this (very old) issue and I _think_ this is the best solution. In the example shared, the values such as `outer_cv <- 5` are not in the call but in the calling environment.

One thing that confuses me a bit is that the *default* for `eval_tidy()` is `rlang::caller_env()` but if I don't explicitly set it like I do here, we still can't find those values.

Anyway here is what we get now:

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(rsample)

run_experiment <- function(all_dataset) {
  outer_cv <- 5
  inner_cv <- 4
  sampling1 <- nested_cv(all_dataset,
                         outside = group_vfold_cv(v = outer_cv, group = "Rep"),
                         inside = group_vfold_cv(v = inner_cv, group = "Rep"))
  
  sampling2 <- nested_cv(all_dataset,
                         outside = group_vfold_cv(v = outer_cv, group = "Rep2"),
                         inside = group_vfold_cv(v = inner_cv, group = "Rep2"))
  
  list(sampling1, sampling2)
}

all_dataset <- matrix(nrow = 50, ncol = 5, 0) %>% as.data.frame()
all_dataset$Rep <- 1:5
all_dataset$Rep2 <- 5:1
run_experiment(tibble(all_dataset))
#> [[1]]
#> # Nested resampling:
#> #  outer: Group 5-fold cross-validation
#> #  inner: Group 4-fold cross-validation
#> # A tibble: 5 × 3
#>   splits          id        inner_resamples         
#>   <list>          <chr>     <list>                  
#> 1 <split [40/10]> Resample1 <group_vfold_cv [4 × 2]>
#> 2 <split [40/10]> Resample2 <group_vfold_cv [4 × 2]>
#> 3 <split [40/10]> Resample3 <group_vfold_cv [4 × 2]>
#> 4 <split [40/10]> Resample4 <group_vfold_cv [4 × 2]>
#> 5 <split [40/10]> Resample5 <group_vfold_cv [4 × 2]>
#> 
#> [[2]]
#> # Nested resampling:
#> #  outer: Group 5-fold cross-validation
#> #  inner: Group 4-fold cross-validation
#> # A tibble: 5 × 3
#>   splits          id        inner_resamples         
#>   <list>          <chr>     <list>                  
#> 1 <split [40/10]> Resample1 <group_vfold_cv [4 × 2]>
#> 2 <split [40/10]> Resample2 <group_vfold_cv [4 × 2]>
#> 3 <split [40/10]> Resample3 <group_vfold_cv [4 × 2]>
#> 4 <split [40/10]> Resample4 <group_vfold_cv [4 × 2]>
#> 5 <split [40/10]> Resample5 <group_vfold_cv [4 × 2]>
```

<sup>Created on 2021-11-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/nest.R
+++ b/R/nest.R
@@ -47,7 +47,6 @@
 #' sum(grepl("Volvo 142E", rownames(inner_assess)))
 #' @export
 nested_cv <- function(data, outside, inside)  {
-  nest_args <- formalArgs(nested_cv)
   cl <- match.call()
 
   boot_msg <-
@@ -60,22 +59,22 @@ nested_cv <- function(data, outside, inside)  {
   outer_cl <- cl[["outside"]]
   if (is_call(outer_cl)) {
     if (grepl("^bootstraps", deparse(outer_cl)))
-      warning(boot_msg, call. = FALSE)
-    outer_cl$data <- quote(data)
-    outside <- eval(outer_cl)
+      rlang::warn(boot_msg)
+    outer_cl <- rlang::call_modify(outer_cl, data = data)
+    outside <- rlang::eval_tidy(outer_cl, env = rlang::caller_env())
   } else {
     if (inherits(outside, "bootstraps"))
-      warning(boot_msg, call. = FALSE)
+      warn(boot_msg)
   }
 
   inner_cl <- cl[["inside"]]
   if (!is_call(inner_cl))
-    stop(
+    abort(
       "`inside` should be a expression such as `vfold()` or ",
-      "bootstraps(times = 10)` instead of a existing object.",
-      call. = FALSE
+      "bootstraps(times = 10)` instead of an existing object.",
     )
   inside <- map(outside$splits, inside_resample, cl = inner_cl)
+  inside <- map(inside, rlang::eval_tidy, env = rlang::caller_env())
 
   out <- dplyr::mutate(outside, inner_resamples = inside)
 
@@ -88,8 +87,7 @@ nested_cv <- function(data, outside, inside)  {
 }
 
 inside_resample <- function(src, cl) {
-  cl$data <- quote(as.data.frame(src))
-  eval(cl)
+  rlang::call_modify(cl, data = as.data.frame(src))
 }
 
 #' @export
--- a/R/nest.R
+++ b/R/nest.R
@@ -48,6 +48,7 @@
 #' @export
 nested_cv <- function(data, outside, inside)  {
   cl <- match.call()
+  env <- rlang::caller_env()
 
   boot_msg <-
     paste0(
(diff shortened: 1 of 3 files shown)


R/nest.R
```diff
@@ -47,7 +47,6 @@
 #' sum(grepl("Volvo 142E", rownames(inner_assess)))
 #' @export
 nested_cv <- function(data, outside, inside)  {
-  nest_args <- formalArgs(nested_cv)
   cl <- match.call()
```
DavisVaughan, line 51: I think I would do an explicit:

```
env <- rlang::caller_env()
```

Then pass `env` through to `eval_tidy()` everywhere you need it

R/nest.R
```diff
@@ -88,8 +87,7 @@ nested_cv <- function(data, outside, inside)  {
 }
 
 inside_resample <- function(src, cl) {
-  cl$data <- quote(as.data.frame(src))
-  eval(cl)
+  rlang::call_modify(cl, data = as.data.frame(src))
 }
```
DavisVaughan, line 91: If you want to avoid needing the second `map()` to call `eval_tidy()`, then you could rewrite this as:

```r
inside_resample <- function(src, cl, env) {
  cl <- rlang::call_modify(cl, data = as.data.frame(src))
  eval_tidy(cl, env = env)
}
```

DavisVaughan (commented):
Regarding the explicit passing through of `caller_env()`, maybe this helps:

(this is also how our `withr::local_*()` functions work)

``` r
# Capture expression, then evaluate it.
# One of these is explicit about the caller_env().

# - caller_env() is evaluated directly in capture_eval1(),
#   so the result of caller_env() is the env that capture_eval1() is called from
capture_eval1 <- function(expr) {
  expr <- rlang::enexpr(expr)
  rlang::eval_tidy(expr, env = rlang::caller_env())
}

# - caller_env() is evaluated inside eval_tidy(),
#   so the result of caller_env() is the env of capture_eval2()
#   (and x is not defined there!)
capture_eval2 <- function(expr) {
  # caller_env() points HERE
  expr <- rlang::enexpr(expr)
  rlang::eval_tidy(expr)
}

# Now lets mimic the user's function
fn1 <- function() {
  # caller_env() points HERE
  x <- 2L
  capture_eval1(x)
}

fn2 <- function() {
  x <- 2L
  capture_eval2(x)
}

# - caller_env() is the fn1() function env, so we find x
fn1()
#> [1] 2

# - caller_env() is the capture_eval2() function env, so we cant find x
fn2()
#> Error in rlang::eval_tidy(expr): object 'x' not found
```

R/nest.R
DavisVaughan: You probably don't even need `eval_tidy()` since we aren't dealing with quosures at all. I think `eval()` might be sufficient. The missing piece was passing through the right `env`

juliasilge:
Thanks @DavisVaughan! 🙌

PR #271, opened 2021-11-18
Return a `repeated_vfold_cv` class object for repeats > 1

### LINKED ISSUE #82: vfold_cv(repeats > 1) should return a `repeated_vfold_cv` class object
Because the column structure changes of the returned tibble, and there are two columns for `id`, it would be nice if the class was different. This will be useful when tuning over a repeated cv object so we make sure and extract both `id` and `id2`.

juliasilge:
Closes #82 

``` r
library(rsample)
folds1 <- vfold_cv(mtcars, v = 10)
folds2 <- vfold_cv(mtcars, v = 10, repeats = 2)

class(folds1)
#> [1] "vfold_cv"   "rset"       "tbl_df"     "tbl"        "data.frame"
class(folds2)
#> [1] "repeated_vfold_cv" "vfold_cv"          "rset"             
#> [4] "tbl_df"            "tbl"               "data.frame"
```

<sup>Created on 2021-11-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I also added in the new-ish `validation_time_split()` to the vctrs tests because I realized we had forgotten it, while working with the vctrs compatibility stuff here.

I hate to tag you again so quickly @DavisVaughan but since this has vctrs compatability work involved, do you mind another review? I hope this one is quick.


DavisVaughan:
Here: https://github.com/tidymodels/rsample/pull/271/files#diff-29a83128120434ed3c2090b36174b5864979e14d2a80682832e532a0c51a9e95L123

I think we might want

```
vfold_cv = vfold_cv(test_data(), v = 10, repeats = 1),
repeated_vfold_cv = vfold_cv(test_data(), v = 10, repeats = 2),
```

R/vfold.R
```diff
@@ -68,9 +69,11 @@ vfold_cv <- function(data, v = 10, repeats = 1,
   strata_check(strata, data)
 
   if (repeats == 1) {
+    cv_subclass <- c("vfold_cv", "rset")
     split_objs <- vfold_splits(data = data, v = v,
                                strata = strata, breaks = breaks, pool = pool)
   } else {
+    cv_subclass <- c("repeated_vfold_cv", "vfold_cv", "rset")
```
DavisVaughan, line 76: In my ideal world, we'd have `vfold_cv(data, v = )` and `repeated_vfold_cv(data, v = , repeats = )`. That way both functions could be type stable. But I guess this is the best we can do since this function is so old.

DavisVaughan (approved):

R/vfold.R
juliasilge: Honestly I don't think it would be super disruptive or bad to do `repeats = deprecated()` and make a new function. It would be pretty straightforward internally.

DavisVaughan: Oooh and what if `repeated_vfold_cv()` returned just a single id column which was a combination of `Repeats<i>_Fold<j>`

``` r
library(rsample)

result <- vfold_cv(mtcars, v = 10, repeats = 2)

# Bad for tune
head(result)
#> # A tibble: 6 × 3
#>   splits         id      id2   
#>   <list>         <chr>   <chr> 
#> 1 <split [28/4]> Repeat1 Fold01
#> 2 <split [28/4]> Repeat1 Fold02
#> 3 <split [29/3]> Repeat1 Fold03
#> 4 <split [29/3]> Repeat1 Fold04
#> 5 <split [29/3]> Repeat1 Fold05
#> 6 <split [29/3]> Repeat1 Fold06

# Good for tune
head(tidyr::unite(result, col = "id", id, id2, sep = "_"))
#> # A tibble: 6 × 2
#>   splits         id            
#>   <list>         <chr>         
#> 1 <split [28/4]> Repeat1_Fold01
#> 2 <split [28/4]> Repeat1_Fold02
#> 3 <split [29/3]> Repeat1_Fold03
#> 4 <split [29/3]> Repeat1_Fold04
#> 5 <split [29/3]> Repeat1_Fold05
#> 6 <split [29/3]> Repeat1_Fold06
```

<sup>Created on 2021-11-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

juliasilge:
We have decided to **not** change the class options or the dataframe return structure, as described in #82

PR #273, opened 2022-02-04
Simulation functions for classification and regression

### LINKED ISSUE #272: Create simulation functions
The caret package has some functions for simulation [like `twoClassSim()`](https://rdrr.io/rforge/caret/man/twoClassSim.html) which would be super helpful for benchmarking, as well as other uses. Let's create some tibble-based, pipeable functions for similar simulation.

@topepo

topepo:
Closes #272


juliasilge (approved):
This is going to such a nice addition, especially for benchmarking 🙌

PR #296, opened 2022-06-10
No more double `vars_select()` 🚫 

### LINKED ISSUE #294: Note: Using an external vector in selections is ambiguous.
## The problem

I'm having a really hard time finding the source or fitting a reprex (sorry! I'll keep trying), but: rsample generates the following message when using `strata` or `group` arguments for the first time:

```
"Message"
"  Note: Using an external vector in selections is ambiguous."
"  i Use `all_of(group)` instead of `group` to silence this message."
"  i See <https://tidyselect.r-lib.org/reference/faq-external-vector.html>."
"  This message is displayed once per session."
```

You can see the message in rsample's CI here:

https://github.com/tidymodels/rsample/runs/6760867450?check_suite_focus=true#step:6:182

But I don't think I understand how to capture whatever generates these messages in a reprex.

juliasilge:
Before this PR, `tidyselect::vars_select()` was called twice inside of `initial_split()`. This generated messages like:

```
"Message"
"  Note: Using an external vector in selections is ambiguous."
"  i Use `all_of(group)` instead of `group` to silence this message."
"  i See <https://tidyselect.r-lib.org/reference/faq-external-vector.html>."
"  This message is displayed once per session."
```

We don't need to `vars_select()` twice, and instead can pass `strata` as `{{ strata }}`. Closes #294

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,14 @@
 
 * Finally removed the `gather()` method for `rset` objects. Use `tidyr::pivot_longer()` instead (#280).
 
+* Changed `initial_split()` to avoid calling tidyselect twice on `strata` (#296). This fix stops `initial_split()` from generating messages like:
+
+```
+  Note: Using an external vector in selections is ambiguous.
+  i Use `all_of(strata)` instead of `strata` to silence this message.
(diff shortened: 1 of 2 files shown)


juliasilge:
This does in fact remove the message from the GH action runs ⬆️

PR #300, opened 2022-06-13
Add options to balance observations between folds

### LINKED ISSUE #298: More control over `group_vfold` behavior when `v < length(unique(group))`
## Feature

(This was originally mentioned to me on Slack by @EmilHvitfeldt as a thought for {spatialsample})

Right now, if `v < length(unique(group))`, I _believe_ that `group_vfold_cv` will randomly combine groups together until each fold has roughly the same number of _groups_. I don't think there's anything wrong with that, but it does mean that if groups are unbalanced it's possible to wind up with pretty dramatically unbalanced folds:

``` r
library(rsample)
data("ames", package = "modeldata")

set.seed(1107)
group_vfold_cv(ames, Neighborhood, 10)
#> # Group 10-fold cross-validation 
#> # A tibble: 10 × 2
#>    splits             id        
#>    <list>             <chr>     
#>  1 <split [2562/368]> Resample01
#>  2 <split [2672/258]> Resample02
#>  3 <split [2727/203]> Resample03
#>  4 <split [2372/558]> Resample04
#>  5 <split [2583/347]> Resample05
#>  6 <split [2366/564]> Resample06
#>  7 <split [2592/338]> Resample07
#>  8 <split [2699/231]> Resample08
#>  9 <split [2900/30]>  Resample09
#> 10 <split [2897/33]>  Resample10
```

<sup>Created on 2022-06-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

It might be nice to have an alternative method for combining groups which attempted to lump the smallest groups together, in a similar way to how `strata` uses the `pool` argument. If we added this in rsample, I might also extend it in spatialsample to combine groups based on geographic proximity.

mikemahoney218:
This PR closes #298 by providing a new argument, `balance`. This argument can be one of either `"groups"`, which is the historical default, or `"observations"`, which will attempt to combine groups so as to balance the number of observations between folds. 

It also renames the test file from `test-group.R` to `test-groups.R` so that `usethis` finds it for `test_active_file`.


juliasilge (approved):
This is so great! Thank you so much 🙌 

I think this is a good thing to get into rsample 1.0.0 as is, without extending to a custom grouping function. We won't be blocked on making changes to this or extending later in the summer.

PR #302, opened 2022-06-14
Initial split printing

topepo:
Instead of analysis/assessment, we'll now get

```r
> library(rsample)
> initial_time_split(mtcars)
<Training/Testing/Total>
<24/8/32>
```


juliasilge (approved):
Thanks! 🙌

PR #303, opened 2022-06-14
Simplify printing

### LINKED ISSUE #292: pretty.<class> documentation can be simplified
As part of https://github.com/tidymodels/spatialsample/pull/20 we decided to remove the documentation for spatialsample `pretty()` methods, given that neither spatialsample or rsample document their `print()` methods and it removes some repetitiveness when adding new `pretty()` methods. 

We thought it might be useful to do the same here, removing most of the roxygen around each `pretty()` method and just exporting them silently instead. For instance, this:

https://github.com/tidymodels/rsample/blob/a4bb47b6bbf0cb79bd17ca456db1af73bf5f11b7/R/labels.R#L94-L98

Becomes this:

https://github.com/tidymodels/spatialsample/blob/main/R/labels.R#L1-L2

mikemahoney218:
This PR closes #292 by removing the roxygen tags (save `@export`) from `pretty` methods, and by moving all `print` and `pretty` methods to a new file `printing.R`. 

I left the print method for rsplits in `rsplit.R`, because that feels like its own _thing_ in a way these other methods don't, but I'm not married to that decision.


juliasilge:
The `gather.rset` was deprecated in rsample and removed in tidyposterior in tidymodels/tidyposterior#53, but that's not on CRAN yet (only on GH).

For the error itself, it looks like all the `pretty` methods got un-exported (check out the NAMESPACE file). We do still need to export those. Can you restart R and then `devtools::document()` again?

mikemahoney218:
Hmm, so we still have the S3 methods registered ( https://github.com/tidymodels/rsample/blob/mike/292/NAMESPACE#L26-L39 ) and `devtools` doesn't export `pretty` separately when we document things this way. That lines up with how things are in spatialsample ( https://github.com/tidymodels/spatialsample/blob/main/NAMESPACE ), since I _think_ they shouldn't be importing S3 methods, but just using the base generic?

mikemahoney218:
I don't _think_ tidyposterior needs to import these, either -- testing against CI in 
https://github.com/mikemahoney218/tidyposterior/actions/runs/2498414288

If you think this change would be OK, I can also open a PR on the tidyposterior repository

juliasilge:
Oh gosh, what is even going on -- I hadn't noticed that all the pretty methods were in the NAMESPACE twice, or that tidyposterior imported pretty methods. 🫠

I did spring cleaning on tidyposterior in March so I hope it's not too painful, but yes, if you can do a PR over there, that would be great. We use the GH version in the pkgdown site here, so that should be OK.

mikemahoney218:
There's a challenge with an orphaned CRAN package giving a warning, but based on CI this change is pretty painless: https://github.com/tidymodels/tidyposterior/pull/59

juliasilge:
I'm running revdepchecks on this branch to see if this is good to go. I can't check tidyposterior no matter what, [because it can't be installed, it looks like](https://github.com/tidymodels/rsample/blob/main/revdep/cran.md).

Issue #207, opened 2021-01-14
more group-based splitting methods


topepo:
It would be good to have an `initial_group_split(data, group, strata, prop)` method that can split the data when there are groups (perhaps patients). The  `strata` option might be difficult when the outcome (or other stratification variable) is not constant within each group. We could also use the median or mode on the stratification variable and use that. 

Similarly, a `mc_group_cv()` function would also be a good idea (using the splitting method as above).

NateMietk:
I second this feature! I have pre-defined testing/training datasets so I cannot use the split_initial function, which in turn does not allow me to utilize the the last_fit function. This would open up the world of tidymodels for me and other uses like me!

juliasilge:
@NateMietk It sounds like you may be interested in a different kind of feature, one that we already support. If you want to create an initial split from your own training/testing data, then you want to use `rsample::make_splits()`. Check out:

- [this Stack Overflow answer](https://stackoverflow.com/a/64004636/5468471)
- [this GitHub issue](https://github.com/tidymodels/rsample/issues/158)

for some examples on how to use it.

hnagaty:
I second this too. I came across a data set that has a grouping variable. I already used `group_vfold_cv()` for the cross validation resamples, and it would be helpful to have a similar approach to the initial split.

rkb965:
This would be a great idea. :) I'm trying to bootstrap with repeated measures data and want all measurements from an individual to be in the same split. It seems like this grouped `initial_split` would enable this, correct? Do y'all have any ideas for work arounds in the meantime? Thanks for all of the incredible work on `tidymodels`!

juliasilge:
Yes @rkb965 in the meantime, you can create a custom split with [`make_splits()`](https://rsample.tidymodels.org/reference/make_splits.html). Check out:

- [this Stack Overflow answer](https://stackoverflow.com/a/64004636/5468471)
- [this GitHub issue](https://github.com/tidymodels/rsample/issues/158)

for some examples on how to use it.

astamm:
Assuming that stratification variable strategies the groups, could something like have any sense:
```
group_initial_split <- function(data, group, prop = 3/4, strata = NULL, breaks = 4, pool = 0.1, ...) {
  strata <- enquo(strata)
  group <- enquo(group)
  data <- nest(data, data = !any_of(c(!!group, !!strata)))
  cell_split <- initial_split(data, prop, !!strata, breaks, pool, ...)
[...]

PR #308, opened 2022-06-21
Refactor group logic

mikemahoney218:
This PR is a first step towards addressing #207 , by starting to factor some of the grouping logic out of `group_vfold_cv` and into its own functions. The broad goal here is to support additional functions (particularly `initial_split` and `mc_cv`) by expanding the `switch` statement with additional `balance` arguments. 

This PR shouldn't add any new features or change any behavior, but just reorganizes the grouping internals so that future changes are less invasive. It doesn't need to get in pre-1.0.0, there's no rush here.


juliasilge (changes requested):
This is a great start to these larger changes! 🙌 

- Can you have the function be `make_groups()`, to parallel `make_strata()`?
- And can you pull that function into a new file `make_groups.R`?

We also should change this filename but I'm not totally sure yet what makes the most sense: `group_vfold.R` vs `vfold_group.R`? If you don't have a strong opinion, let's leave that renaming to when we add a new grouped method.

mikemahoney218:
I was debating asking if we should move it into vfold_cv, and then have future methods also in the same file as their non group methods. Otherwise, not sure what file names would be best!

juliasilge:
Oh, I think I like the idea of each grouped function being in the same file as its ungrouped counterpart! 👍 No need for that here, but for future PRs.

juliasilge:
Should we document and export this like `make_strata()`? 
https://rsample.tidymodels.org/reference/make_strata.html
I think the reason to do this would be so we can `@inheritParams` nicely, and do something like "see [make_groups()]".

mikemahoney218:
I think I'd personally like to _not_ export this if we're merging it pre-1.0.0, because I think we might want to change how the `switch` functions work once we're also providing `group_initial_split` and `group_mc_cv`. 

I haven't thought about this enough to take a stab at implementing it yet, but we also discussed in #300 possibly expanding grouping to take what would effectively be a custom `balance_*` function, which would _probably_ require passing `...` to the `balance_*` functions. That means we might also change how things get passed to `make_groups`, so that different functions can use different arguments -- for instance, we won't need `v` for `group_initial_split`. But I haven't thought enough about what that could look like to even float an implementation yet, and I imagine we'll want to have a bigger discussion about exactly how this all should work in the nearish future.

All that to say, more than happy to document here (and I think we can `@inheritParams` from unexported functions), but I think it'd be nice to wait on exporting this for another version until we get grouping fully implemented.

juliasilge (approved):
This is a great beginning, and your plan to not export makes good sense. 👍

PR #310, opened 2022-06-23
Remove balance

### LINKED ISSUE #309: Should we pull `balance` out of `group_vfold_cv()` for 1.0?
I'm losing a bit of my confidence that `balance`, as it currently exists in `group_vfold_cv()`, is going to be what we want to support long-term. Given that, should we pull the `balance` option from this coming release, and then add it back when we implement the rest of our grouping?

There are two main challenges with the current approach:

1. Right now, the two options (`groups` and `observations`) don't need to incorporate any arguments other than an index of the data, a corresponding vector of group IDs, and `v`. As we expand to `mc_cv()` and `initial_split()`, we're going to need to find a way to incorporate `prop` as an argument, and I think we want to incorporate `strata` as well (to assign groups to folds based on stratified variables). Having a character vector of options for `balance` makes dealing with this a bit challenging without introducing complex dependencies between arguments.
2. `balance = "observations"` is deterministic, and I don't know if we _like_ that. At the very least, it means that bringing it to things like `mc_cv()` is difficult, since we very much _don't_ want those to be deterministic samples.

I've got some thoughts about how to approach this, but I don't think they're going to make it by the 1.0 release. How would we feel about not introducing `balance` in this release, and instead including it with a broader `group` release?

mikemahoney218:
This PR fixes #309 .

I _believe_ this catches everything -- balance is no longer an argument to `group_vfold_cv()` or `group_vfold_splits()` or `make_groups()`, I removed `balance_observations()`, there's no more tests about balance arguments, and the item has been removed from the NEWS.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,8 +16,6 @@
 
 * Added better printing methods for initial split objects.
 
-* Added a new `balance` option to `group_vfold_cv()` to balance folds either by the number of groups or the number of observations (@mikemahoney218, #300).
-
 # rsample 0.1.1
 
 * Updated documentation on stratified sampling (#245).
--- a/R/groups.R
+++ b/R/groups.R
@@ -22,10 +22,9 @@
 #'
 #' set.seed(123)
 #' group_vfold_cv(Sacramento, group = city, v = 5)
-#' group_vfold_cv(Sacramento, group = city, v = 5, balance = "observations")
 #'
 #' @export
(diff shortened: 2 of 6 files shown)


juliasilge (approved):
This looks great -- thank you! I am going to restart the release process today for 1.0.0 and then we can move forward with newer features like this.

mikemahoney218:
I promise I won't have any last minute requests :laughing:

PR #311, opened 2022-06-24
Add new balance implementation

mikemahoney218:
This PR starts to replace what we removed in #310 (which undid #300), adding in a more flexible interface for balancing groups.

The current implementation has methods for having roughly equal numbers of _groups_ and of _observations_ across folds, using two new helper functions (`balance_groups()` and `balance_observations()`) which take no arguments. The observations implementation is now stochastic and so changes across runs, but as a consequence no longer balances precisely; I think we prefer that.

My intent is that we can then spread this to incorporate `balance_prop()` (which is like `balance_observations()`, but won't assign each group to a fold, for `mc_cv()`) and then down the line `balance_strata()`. 

``` r
library(ggplot2)
library(rsample)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union


data(ames, package = "modeldata")
set.seed(123)

analysis_sd <- function(v, balance) {
  group_vfold_cv(ames, Neighborhood, v, balance = rlang::exec(get(balance)))$splits |> 
    purrr::map_dbl(~ nrow(analysis(.x))) |> 
    sd()
}

out <- tidyr::crossing(
  idx = seq_len(100),
  v = c(2, 5, 10, 15),
  balance = c("balance_groups", "balance_observations")
)

out |> 
  mutate(sd = purrr::pmap_dbl(
    list(v, balance),
    analysis_sd
  )) |> 
  ggplot(aes(sd, fill = balance)) + 
  geom_histogram(alpha = 0.6, color = "black", size = 0.3) + 
  facet_wrap(~ v) + 
  theme_minimal() + 
  labs(title = "sd() of nrow(analysis) by balance method")
#> `stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
```

<sup>Created on 2022-06-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


R/make_groups.R
```diff
@@ -27,12 +49,90 @@ make_groups <- function(data, group, v) {
...
+#' @return A list with three elements: `fn`, the function which will actually
+#'  produce groups for the resampling method, `args`, a list of
+#'  any balancing-method specific arguments that `fn` requires, and `type`,
+#'  indicating what balancing method is being used.
+#'
+#' @rdname balance
+#' @export
+balance_groups <- function(...) {
+  rlang::check_dots_empty()
+  list(
```
mikemahoney218, line 68: The user-facing, exported `balance_*` functions return a list of the _actual_ grouping engine and the arguments it requires, so that we can pass things created in `make_groups()` (specifically, `data_ind`) to the grouping engines and don't need to expose the weirder implementation details to the user.

It's a weird way to do this, but I _think_ it makes the most sense. If we refactored so that `data_ind` was created inside the `balance_*` functions, not `make_groups()`, then it'd be a lot harder for other packages (for instance, `spatialsample` to add balancer functions.

R/make_groups.R
```diff
@@ -27,12 +49,90 @@ make_groups <- function(data, group, v) {
...
+    fn = make_balance_groups,
+    args = list(),
+    type = "balance_groups"
```
mikemahoney218, line 71: My thought is that we can use this `type` variable to restrict some functions to only use certain balancers -- for instance, to make sure that `group_vfold_cv()` doesn't use `balance_prop()`, which won't assign each value to a fold.

mikemahoney218:
I opened #312 to highlight what I mean by this being "extendable", showing how this might work for `mc_cv()`.

R/make_groups.R
hfrick: It does feel a little weird :sweat_smile: While we obviously don't want to build anything that's terrible hard to extend, I think it's worth considering if there are ways to keep the interface a little less "weird" (to stay with technical terms here) for the users at the price of making it a little bit more tedious for the developers. 

The draft PR for `group_mc_cv()` was a good idea! I think the current API of `group_mc_cv(data, group, times = 25, balance = balance_prop(prop = 3 / 4), ...)` holds a bit much at once with the argument `prop` inside of the `balance` argument.

Is your point that in spatialsample you want to be able to pass a function for balancing to `group_vfold_cv()`? If so, would a similar approach as to the `clustering_function` argument to `spatial_clustering_cv()` make sense? I.e., users could pick a type of balancing via a string but passing a function also works?

mikemahoney218: Yeah, this all makes sense. I think I was concerned about how we'd deal with stratification, in order to avoid strata doing different things inside the "same" functions based on the balancing method. I'm not as concerned about that after the weekend.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # rsample (development version)
 
+* Added functions to control how grouped resampling combines groups. Use `balance = balance_groups()` to assign (roughly) the same number of groups to each fold, or `balance = balance_observations()` to assign (roughly) the same number of observations to each fold.
```
hfrick, line 3:
```suggestion
* Added functions to control how grouped resampling combines groups. Use `balance = "groups"` to assign (roughly) the same number of groups to each fold, or `balance = "observations"` to assign (roughly) the same number of observations to each fold.
```

hfrick (approved):

juliasilge (approved):
This is a great initial approach for this! 🙌

PR #316, opened 2022-06-28
Add `group_bootstraps()`

mikemahoney218:
This PR is based off of #315 . 

This PR fixes #207 (and we're going to open a new issue for stratification with grouping, to get input from a wider audience).

This PR adds `group_bootstraps()`, patterned off `group_mc_cv()`, in order to provide a way to bootstrap with defined groups. In order to do that, it adds an argument `replace` to `balance_prop()`, which is `TRUE` for `group_bootstraps()` and `FALSE` for `group_mc_cv()`. 

``` r
library(rsample)

set.seed(2)
dat1 <- tibble::tibble(a = 1:20, b = letters[1:20], c = rep(1:4, 5))

group_bootstraps(dat1, c, times = 2)$splits[[1]] |> 
  analysis()
#> # A tibble: 20 × 3
#>        a b         c
#>    <int> <chr> <int>
#>  1     1 a         1
#>  2     1 a         1
#>  3     2 b         2
#>  4     4 d         4
#>  5     5 e         1
#>  6     5 e         1
#>  7     6 f         2
#>  8     8 h         4
#>  9     9 i         1
#> 10     9 i         1
#> 11    10 j         2
#> 12    12 l         4
#> 13    13 m         1
#> 14    13 m         1
#> 15    14 n         2
#> 16    16 p         4
#> 17    17 q         1
#> 18    17 q         1
#> 19    18 r         2
#> 20    20 t         4

group_bootstraps(dat1, c, times = 2)$splits[[1]] |> 
  analysis()
#> # A tibble: 20 × 3
#>        a b         c
#>    <int> <chr> <int>
#>  1     1 a         1
#>  2     2 b         2
#>  3     2 b         2
#>  4     3 c         3
#>  5     5 e         1
#>  6     6 f         2
#>  7     6 f         2
#>  8     7 g         3
#>  9     9 i         1
#> 10    10 j         2
#> 11    10 j         2
#> 12    11 k         3
#> 13    13 m         1
#> 14    14 n         2
#> 15    14 n         2
#> 16    15 o         3
#> 17    17 q         1
#> 18    18 r         2
#> 19    18 r         2
#> 20    19 s         3

group_bootstraps(dat1, c, times = 2)$splits[[1]] |> 
  assessment()
#> # A tibble: 5 × 3
#>       a b         c
#>   <int> <chr> <int>
#> 1     2 b         2
#> 2     6 f         2
#> 3    10 j         2
#> 4    14 n         2
#> 5    18 r         2

[...]

PR #319, opened 2022-06-29
Add reverse_splits

### LINKED ISSUE #284: Reversing analysis and assessment splits
Hi,

I use the `group_vfold_cv()` function a lot, but I have recently been needing to reverse the analysis and assessment splits. For example, lets say you have 3 groups, `group_vfold_cv()` would create three folds, whereby within each fold 2 groups are used for training, and one is held aside for testing. I would like to do the opposite, where one group is used for training, and the remaining groups are used for testing.

I don't think this is possible using `make_splits()`, unless I'm just doing it incorrectly. Would you ever consider supporting this resampling scheme? If not, is it possible for me to create these kinds of splits using the existing `rsample` machinery?

mikemahoney218:
This PR fixes #284 by adding a new utility function to "reverse" the analysis and assessment splits.

``` r
library(rsample)

(starting_splits <- vfold_cv(mtcars, v = 3))
#> #  3-fold cross-validation 
#> # A tibble: 3 × 2
#>   splits          id   
#>   <list>          <chr>
#> 1 <split [21/11]> Fold1
#> 2 <split [21/11]> Fold2
#> 3 <split [22/10]> Fold3

reverse_splits(starting_splits)
#> #  3-fold cross-validation 
#> # A tibble: 3 × 2
#>   splits          id   
#>   <list>          <chr>
#> 1 <split [11/21]> Fold1
#> 2 <split [11/21]> Fold2
#> 3 <split [10/22]> Fold3

reverse_splits(starting_splits$splits[[1]])
#> <Analysis/Assess/Total>
#> <11/21/32>

reverse_splits(1)
#> Error in `reverse_splits()`:
#> ! `x` must be either an `rsplit` or an `rset` object

reverse_splits(starting_splits, "no dots allowed!")
#> Error in `reverse_splits()`:
#> ! `...` must be empty.
#> ✖ Problematic argument:
#> • ..1 = "no dots allowed!"
#> ℹ Did you forget to name an argument?
```

<sup>Created on 2022-06-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


R/compat-vctrs-helpers.R
```diff
@@ -109,7 +109,7 @@ rset_reconstructable <- function(x, to) {
 test_data <- function() {
   data.frame(
     x = 1:50,
-    y = rep(c(1, 2), each = 25),
+    y = rep(seq.int(10), each = 5),
```
mikemahoney218, line 112: Only having two groups here made `group_mc_cv()` error due to having 0 data in the assessment set. I _also_ did a version of this in #316 to deal with the same issue there (to get both PRs to pass CI).

juliasilge:
@mattwarkentin would you like to take a look at this?

R/misc.R
```diff
@@ -150,3 +150,63 @@ split_unnamed <- function(x, f) {
   }
   res
 }
+
+#' Reverse the analysis and assessment sets
+#'
+#' This functions "swaps" the analysis and assessment sets of either a single
+#' `rsplit` or all `rsplits` in the `split` column of an `rset` object.
```
mattwarkentin, line 157: I think it is `splits` column

mattwarkentin: Also maybe `rsplit`s for the second mention:

> This function "swaps" the analysis and assessment sets of either a single `rsplit` or all `rsplit`s in the `splits` column of an `rset` object.

mattwarkentin:
Oh, one more thought. Have you tested this function with an object from `permutations()`. Splits for a `permutations` object have no assessment set by design, so not sure how this should be handled.

mattwarkentin:
Related: I am now seeing that the error throw with calling `assessment()` on a `permutations` object is not correct:
``` r
library(rsample)

p <- permutations(mtcars, cyl)

assessment(p$splits[[1]])
#> Error in `as.data.frame()`:
#> ! There is no assessment data set for an `rsplit` object with class `rsplit`.
```

It should read something like:
> #> ! There is no assessment data set for an `rsplit` object from a `permutations` set.

Or something along those lines...

mikemahoney218:
> Related: I am now seeing that the error thrown when calling `assessment()` on a `permutations` object is not correct...

Opened #321 to track that one :smile:

juliasilge (approved):
Thank you so much! 🙌

PR #322, opened 2022-06-29
Use the non-rsplit class in messages 

### LINKED ISSUE #321: Calling `assessment()` on a `permutations` gives a confusing error
Related: I am now seeing that the error thrown when calling `assessment()` on a `permutations` object is not correct:
``` r
library(rsample)

p <- permutations(mtcars, cyl)

assessment(p$splits[[1]])
#> Error in `as.data.frame()`:
#> ! There is no assessment data set for an `rsplit` object with class `rsplit`.
```

It should read something like:
> #> ! There is no assessment data set for an `rsplit` object from a `permutations` set.

Or something along those lines...

_Originally posted by @mattwarkentin in https://github.com/tidymodels/rsample/pull/319#issuecomment-1170492817_

mikemahoney218:
This PR fixes #321 by changing which class is used in the message when calling `assessment()` on an object with a `col_id` slot.

I am 95% sure this only impacts permutations; `grep` suggests that's the only place that `col_id` gets set:

```
❯ grep col_id R/*
R/permutations.R:  col_id <- tidyselect::eval_select(permute, data)
R/permutations.R:  if (identical(length(col_id), 0L)) {
R/permutations.R:  } else if (identical(length(col_id), ncol(data))) {
R/permutations.R:      x$col_id <- col_id
R/permutations.R:    col_id = col_id
R/printing.R:  shuffled_cols <- paste(names(attr(x, "col_id")), collapse = ", ")
R/rsplit.R:    if (!is.null(x$col_id)) {
R/rsplit.R:        x$data[as.integer(x, data = data, ...), x$col_id, drop = FALSE]
R/rsplit.R:      x$data[, x$col_id] <- permuted_col
```


juliasilge (approved):
Yeah, it looks to me like this is right. 👍

PR #326, opened 2022-06-30
Auto-generate vctrs

mikemahoney218:
This PR adds a file, `generate_vctrs.R`, which automatically generates the vctrs compatibility functions needed for functions in `rset_subclasses`. It removes a little bit of the overhead involved in adding new functions and subclasses -- rather than figuring out what needs to be updated, adding a function to `rset_subclasses` will result in the tests and the functions being automatically generated. 

I totally get if we don't want to go down this road, just feels like it might be a bit easier than the current method we update these.

It requires R > 4.1 and is in .Rbuildignore.


juliasilge:
The recipes package has something like this and puts it in `/inst`:
https://github.com/tidymodels/recipes/blob/main/inst/boilerplate.R
That's probably the best place to put it.

We would probably also want to [add this as a release question](https://r-pkgs.org/release.html#release-submission).

What do you think about this @DavisVaughan, as our resident vctrs expert? Seems helpful to me, since we have been known to forget to do this, especially in contributor PRs.

inst/generate_vctrs.R
```diff
@@ -0,0 +1,74 @@
...
+#' @export
+vec_cast.data.frame.CHANGEME <- function(x, to, ..., x_arg = "", to_arg = "") {
+  df_cast(x, to, ..., x_arg = x_arg, to_arg = to_arg)
+})"
+
+generated_template <- vapply(
+  names(rset_subclasses),
+  function(x) {
+    # glue has weird edge cases when trying to interpolate into R code
+    gsub("CHANGEME", x, template)
```
DavisVaughan, line 62: I think you should look into the whisker package for this, which is what usethis uses and is our standard way to do this kind of interpolation, I think

https://github.com/r-lib/usethis/blob/7c8e0049a1e40e6dcabbde069bb29576215a11b6/R/template.R#L62
https://github.com/r-lib/usethis/blob/main/inst/templates/packagename-data-prep.R

DavisVaughan (commented):
I think this is a nice idea, but I would encourage you to generate 1 file per rset-subclass. So you'd have something like:

```
compat-vctrs-bootstraps.R
compat-vctrs-v_fold.R
```

and so on.

I think that will be better because I can imagine having a new subclass in the future where we need to adjust the default cast/ptype2 methods that this will generate for us, and that will be easier to do if we have them in separate files (otherwise rerunning this generator function will overwrite our custom adjustments).

So you'd have a function that writes out 1 file and then you'd call it N times, like:

```r
generate_vctrs_template <- function(class) {
  # do interpolation and write to file
}

generate_vctrs_template("bootstraps")
generate_vctrs_template("v_fold")
# Special case where we started with the template and modified the defaults, don't overwrite
# generate_vctrs_template("special_case")
```

mikemahoney218:
Changed things so that we're generating separate files (which I've prefixed with `z` so they all sort at the bottom). Right now, I've still got a loop creating a file for each of `rset_subclasses`, so that we only need to update one place and get all the compatibility helpers and tests we need. I think if we need to adjust behavior for a new subclass in the future it'd be easy to `setdiff` the vector it iterates through, so that getting the standard compatibility helpers becomes opt-out rather than opt-in.

PR #330, opened 2022-07-01
Add repeats argument (#327)

### LINKED ISSUE #327: `group_vfold_cv()` should have a `repeats` argument
## Feature

As mentioned by @mattwarkentin , `group_vfold_cv()` is currently the only function that's not equivalent to its non-group version (other than stratification), because right now `group_vfold_cv()` does not have a `repeats` argument.

I think it would be nice for `group_vfold_cv()` to have a `repeats` argument. That said, I think we should probably throw an error in _both_ functions when `v == max_v` (either `nrow(data)` or `length(unique(data$group))`), because that means you're going to have identical repeats, which strikes me as undesirable.

mikemahoney218:
I accidentally pushed this to main :facepalm: I reverted immediately (`git reset --soft HEAD~1 && git push -f`) and _believe_ things are fine, but sorry about that.

That said! This PR fixes #327 by:

+ Adding a `repeats` argument to `group_vfold_cv()`
+ Adding an error to _both_ `group_vfold_cv()` and `vfold_cv()` when `repeats` is used with `v == max_v`

``` r
library(rsample)

vfold_cv(mtcars, v = 2, repeats = 2)
#> #  2-fold cross-validation repeated 2 times 
#> # A tibble: 4 × 3
#>   splits          id      id2  
#>   <list>          <chr>   <chr>
#> 1 <split [16/16]> Repeat1 Fold1
#> 2 <split [16/16]> Repeat1 Fold2
#> 3 <split [16/16]> Repeat2 Fold1
#> 4 <split [16/16]> Repeat2 Fold2
vfold_cv(mtcars, v = nrow(mtcars), repeats = 2)
#> Error in `vfold_cv()`:
#> ! Repeated resampling when `v` is 32 would create identical resamples

group_vfold_cv(mtcars, cyl, v = 2, repeats = 2)
#> # Group 2-fold cross-validation 
#> # A tibble: 4 × 3
#>   splits          id      id2      
#>   <list>          <chr>   <chr>    
#> 1 <split [11/21]> Repeat1 Resample1
#> 2 <split [21/11]> Repeat1 Resample2
#> 3 <split [14/18]> Repeat2 Resample1
#> 4 <split [18/14]> Repeat2 Resample2
group_vfold_cv(mtcars, cyl, v = length(unique(mtcars$cyl)), repeats = 2)
#> Error in `group_vfold_cv()`:
#> ! Repeated resampling when `v` is 3 would create identical resamples
group_vfold_cv(mtcars, cyl, v = NULL, repeats = 2)
#> Error in `group_vfold_cv()`:
#> ! Repeated resampling when `v` is `NULL` would create identical resamples
```

<sup>Created on 2022-07-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * Added arguments to control how `group_vfold_cv()` combines groups. Use `balance = "groups"` to assign (roughly) the same number of groups to each fold, or `balance = "observations"` to assign (roughly) the same number of observations to each fold.
 
+* Added a `repeats` argument to `group_vfold_cv()` (#330).
+
 * Added new functions for grouped resampling: `group_mc_cv()` (#313), `group_initial_split()` and `group_validation_split()` (#315), and `group_bootstraps()` (#316).
 
 * Added a new function, `reverse_splits()`, to swap analysis and assessment splits (#319, #284).
--- a/R/vfold.R
+++ b/R/vfold.R
@@ -192,7 +192,7 @@ group_vfold_cv <- function(data, group = NULL, v = NULL, balance = c("groups", "
   } else {
     if (is.null(v)) {
       rlang::abort(
-        glue::glue("Repeated resampling when `v` is `NULL` would create identical resamples")
+        "Repeated resampling when `v` is `NULL` would create identical resamples"
       )
     }
     if (v == length(unique(getElement(data, group)))) {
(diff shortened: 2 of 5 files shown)


juliasilge (approved):
Thank you! 🙌

PR #332, opened 2022-07-01
Set seed when generating rset_subclasses

### LINKED ISSUE #331: `group_bootstraps()` sometimes errors with 0 rows in `compat-vctrs-helpers.R`
See here for an example where this is failing:
https://github.com/tidymodels/rsample/runs/7155341170?check_suite_focus=true#step:6:134

mikemahoney218:
I believe this should fix #331 by making sure we're always generating the same splits. 

The error in `group_bootstraps()` is being thrown on purpose, because there's not enough unique groups in `test_data()`. Because the data inside `rset_subclasses` is only for testing purposes, I think it's fine to just always explicitly use a seed that we know works.


mikemahoney218:
@juliasilge I think this one will solve the CI issues

juliasilge:
Since we are going to use withr only in the `delayedAssign()` and the tests for shuffling, can it be in Suggests?

mikemahoney218:
Moving it to Suggests and redocumenting gives me this on check:

```
* installing *source* package ‘rsample’ ...
** using staged installation
** R
** inst
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
*** copying figures
** building package indices
** installing vignettes
** testing if installed package can be loaded from temporary location
** testing if installed package can be loaded from final location
Error in loadNamespace(x) : there is no package called ‘withr’
Calls: <Anonymous> ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted
ERROR: loading failed
* removing ‘/tmp/RtmpSDudSe/rsample.Rcheck/rsample’
```

So I _believe_ it needs to be in Imports (but could be missing something!).

mikemahoney218:
Oh, should I open an issue to add `R-CMD-check-hard` to this repo? Realizing that we don't actually check for noSuggests right now.

juliasilge:
I definitely know the melted brain 🫠 feeling. Yes, let's add the `check-hard` action here as well so we don't mess something up.

juliasilge (approved):
This should do it, I think 🙌

PR #336, opened 2022-07-05
Add noSuggests check

mikemahoney218:
This PR adds the R-CMD-check-hard action, to ensure that we'll pass noSuggests checks going forward. It also includes code changes to make this new CI pass. Most of these are pretty minimal, skipping tests or examples without suggested packages; the most invasive is that `reg_intervals()` now calls `rlang::check_installed()` at its start.


mikemahoney218:
I didn't add a NEWS item because I don't _think_ this is user-facing, but happy to toss something in if there's something that's worth including.

juliasilge (approved):

PR #339, opened 2022-07-05
Update docs for `group_vfold_cv()`

### LINKED ISSUE #324: `group_loo_cv()` and `group_vfold_cv()`
Hi @mikemahoney218 and @juliasilge,

Super excited for all of the great work @mikemahoney218 has been doing to boost `{rsample}` group sampling support. 

I was thinking about this last night...it seems to me that the current default implementation of `group_vfold_cv()` is really the `group_*` counterpart to `loo_cv()`. With `v = NULL` (the default for `group_vfold_cv()`), you get group LOO-CV sampling. Over time this may be confusing as users may grow to expect the `group_*` version of something to return a sampling pattern similar to the default for its non-grouped sibling.

I am wondering if it makes sense to make the following changes to achieve a logically consistent API for grouped and non-grouped pairs:

- Create `group_loo_cv()` based on the current default implementation of `group_vfold_cv(v = NULL)`

With no changes to `group_vfold_cv()`, this *could* be as simple as:
```r
group_loo_cv <- function(data, group, ...) {
  group_vfold_cv(data, group, v = NULL, ...)
}
```
That way users will find a `group_*` implementation of all relevant standard sampling methods which I think helps the mental model for the grouped verisons.

- However, I wonder if its worth possibly changing the default value of `v` for `group_vfold_cv()` to be equivalent to that of `vfold_cv()` (i.e. `v = 10`) and also adding a `repeats = 1` argument. I think over time people will come to expect that the `group_*` implementation is otherwise equivalent to the non-grouped version, except for the grouping. For `group_vfold_cv()` this currently isn't the case, where the default implementation is actually returning grouped `loo_cv()`. I realize LOO is really a special case of standard v-fold CV and users can opt in to `vfold_cv()`-like sampling, but its not currently the default which may cause mental friction.

I just think for consistency it makes sense for `group_vfold_cv()` to have a signature like:
```r
group_vfold_cv(data, group, v = 10, repeats = 1, ...) {
  ... 
}
```

This would be a behavioural change that could throw a warning/message for some time and then eventually remove the warning/message. I think I just like the idea of `group_*` being otherwise equivalent sampling schemes to their non-group counterparts.

Thoughts?

mattwarkentin:
Closes #324

Related to https://github.com/tidymodels/rsample/issues/324#issuecomment-1172489449

### DIFF
--- a/R/vfold.R
+++ b/R/vfold.R
@@ -151,13 +151,14 @@ vfold_splits <- function(data, v = 10, strata = NULL, breaks = 4, pool = 0.1) {
 #'  on some grouping variable (which may have more than a single row
 #'  associated with it). The function can create as many splits as
 #'  there are unique values of the grouping variable or it can
-#'  create a smaller set of splits where more than one value is left
+#'  create a smaller set of splits where more than one group is left
 #'  out at a time. A common use of this kind of resampling is when you have
 #'  repeated measures of the same subject.
 #'
 #' @inheritParams vfold_cv
-#' @param v The number of partitions of the data set. If left as `NULL`, `v`
-#'  will be set to the number of unique values in the group.
+#' @param v The number of partitions of the data set. If left as `NULL` (the
+#'   default), `v` will be set to the number of unique values in the grouping
+#'   variable, creating "leave-one-group-out" splits.
 #' @param balance If `v` is less than the number of unique groups, how should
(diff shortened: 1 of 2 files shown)


juliasilge (approved):
Thank you so much @mattwarkentin for this and for your thoughtful contributions overall! 🌟

PR #343, opened 2022-07-13
Force sorting from vec_count (fixes #342)

### LINKED ISSUE #342: Identical seeds produce different results across sessions
@mikemahoney218 this is awesome and just what I need, thank you! One thing I noticed though, is that set.seed doesn't seem to be applied to the splits? 

Is it possible to have set.seed allow the group_initial_split() to separate things in the same way?

Here is an example. It does seem to stay the same when repeating the process in the same session, but if you restart R studio you'll see different colors each time. 

```
library(tidyverse)
library(rsample)

df <- starwars %>% 
  mutate(name = factor(name))

set.seed(3332)
group_split <- group_initial_split(df, group = name)
group_train <- training(group_split)
group_test <- testing(group_split)

group_train %>% select(mass, name) %>% mutate(group = "train") %>% 
  bind_rows(group_test %>% select(mass, name) %>% mutate(group = "test")) %>% 
  ggplot(aes(mass, name, color = group))+
  geom_point()
```
Thanks

_Originally posted by @Jeffrothschild in https://github.com/tidymodels/rsample/issues/207#issuecomment-1182754852_

mikemahoney218:
This PR fixes #342 by:

+ Forcing `vec_count()` to sort by key (we had been using the default sort)
+ Testing for consistent output from grouped resampling functions across CI runs

### DIFF
--- a/R/make_groups.R
+++ b/R/make_groups.R
@@ -77,7 +77,7 @@ balance_observations <- function(data_ind, v, ...) {
   n_obs <- nrow(data_ind)
   target_per_fold <- 1 / v
 
-  freq_table <- vec_count(data_ind$..group)
+  freq_table <- vec_count(data_ind$..group, sort = "key")
   freq_table <- freq_table[sample.int(nrow(freq_table)), ]
   freq_table$assignment <- NA
   freq_table$assignment[seq_len(v)] <- seq_len(v)
@@ -115,7 +115,7 @@ balance_prop <- function(prop, data_ind, v, replace = FALSE, ...) {
   }
   n_obs <- nrow(data_ind)
 
-  freq_table <- vec_count(data_ind$..group)
+  freq_table <- vec_count(data_ind$..group, sort = "key")
 
   n <- nrow(freq_table)
   # If sampling with replacement,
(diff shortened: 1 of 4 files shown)


R/make_groups.R
```diff
@@ -77,7 +77,7 @@ balance_observations <- function(data_ind, v, ...) {
   n_obs <- nrow(data_ind)
   target_per_fold <- 1 / v
 
-  freq_table <- vec_count(data_ind$..group, sort = "key")
+  freq_table <- vec_count(data_ind$..group, sort = "location")
```
mikemahoney218, line 80: This PR was originally using `sort = "key"`, but I think that means results might differ across system locales -- I think `sort = "location"` is safer here.

mikemahoney218:
Looks like this will be patched in dev vctrs soon: 
https://github.com/r-lib/vctrs/pull/1589

I'd personally still like to merge this PR, so that our current `main` branch is functional and our release isn't tied to a new release of vctrs. But if we'd rather just bump the dependency, that would also solve the problem.

hfrick (approved):

PR #352, opened 2022-08-01
Add `clustering_cv()`

### LINKED ISSUE #351: Support non-spatial blocking?
Someone brought this up in a workshop -- there's a lot of places that blocked CV can be useful (the Roberts 2017 paper discusses several). It might be useful to support blocking (temporal, phylogenetic) and non-spatial clustering (someone provides a distance matrix, we use a knockoff of spatial_clustering_cv to break it into folds).

mikemahoney218:
This PR fixes #351 by adding `clustering_cv()` for partitioning data based on non-spatial clusters. It extends and simplifies `spatialsample::spatial_clustering_cv()`, adding a `distance_function` argument but removing a bunch of data frame versus sf special casing. It also fixes a number of latent bugs in the spatialsample version, and as a result now works with `reshuffle_rset()`.

It will also be pretty easy to wrap in spatialsample -- my intent is to hardcode `distance_function` to use `sf::st_distance()`, and then approach this using `posthoc_buffer_rset()` just like the vfold wrappers:

``` r
set.seed(123)
data(Smithsonian, package = "modeldata")
smithsonian_sf <- sf::st_as_sf(
  Smithsonian, 
  coords = c("longitude", "latitude"), 
  crs = 4326
)

rsample::clustering_cv(
  smithsonian_sf,
  geometry,
  v = 5,
  distance_function = function(data) {
    as.dist(sf::st_distance(data))
  }
)
#> # 5-cluster cross-validation 
#> # A tibble: 5 × 2
#>   splits         id   
#>   <list>         <chr>
#> 1 <split [18/2]> Fold1
#> 2 <split [19/1]> Fold2
#> 3 <split [18/2]> Fold3
#> 4 <split [16/4]> Fold4
#> 5 <split [9/11]> Fold5
```

<sup>Created on 2022-08-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


juliasilge:
Am I correct in remembering that we want to do an rsample release _before_ merging this in? So the grouping approaches get to CRAN and you can do the blog post without us making this change?

mikemahoney218:
Depends how we feel about this PR! I'd like to do an rsample release and get the blog post out next week, if possible. CRAN submissions are closed this week, so we've got some time there. If we think this is too big a PR to include in the new release _right_ after merging, then I'm more than happy for us to hold it back until the next version. But if we think that we'd be happy with it by the time CRAN submissions re-open, I think it could sneak into this release.

R/clustering.R
```diff
@@ -0,0 +1,142 @@
...
+#' - `dists`, a [stats::dist()] object with distances between data points
+#' - `v`, a length-1 numeric for the number of folds to create
+#' - `...`, to pass any additional named arguments to your function
+#'
+#' The function should return a vector of cluster assignments of length
+#' `nrow(data)`, with each element of the vector corresponding to the matching
+#' row of the data frame.
+#'
+#' @param data A data frame to split into folds.
+#' @param vars A vector of variable names to use to cluster the data.
```
juliasilge, line 32:
```suggestion
#' @param vars A vector of bare variable names to use to cluster the data.
```

juliasilge (approved):
This looks super nice! 🙌 

Are you able to redocument easily @mikemahoney218 for that tiny change? If not, I can do it.

mikemahoney218:
Redocumented :smile:

PR #353, opened 2022-08-02
Fix typo

mikemahoney218:
This is causing a hyperlink to appear in the NEWS file to `break`, which was not intentional :laughing:


mikemahoney218:
:facepalm: this will pass CI if #352 is merged. It doesn't introduce new bugs; the errors are due to the new version of pillar.

juliasilge (approved):
Thank you! 🙌

Issue #317, opened 2022-06-28
Stratification in grouped resampling


mikemahoney218:
## Feature

As part of closing #207, we've recently implemented a number of grouping functions, with `group_mc_cv()` (#313), `group_initial_split()` and `group_validation_split()` (#315), and `group_bootstraps()` (#316). 

Right now, none of these functions support stratification -- which would be useful if, for instance, you had repeated measurements of a number of patients and needed to stratify by outcome. We haven't included this partially so that we could implement grouped resampling quickly, but also because we aren't exactly sure what people would expect stratification to _do_ when resampling by groups. Specific questions include:

+ How should strata be determined when the stratification variable isn't constant within a group? Median, mode, user-provided functions? What's a good default option?
+ What rules can we use to determine when a (group x strata) needs to be pooled with others? 

If anyone has any thoughts on what they'd expect stratification to do in grouping functions, let us know here!

juliasilge:
Posted on [RStudio Community](https://community.rstudio.com/t/stratification-in-grouped-resampling/141071) and [Twitter](https://twitter.com/juliasilge/status/1542142791681187840)

civilstat:
Thank you for providing these grouping functions!  
As for stratification, I'd be interested in two use cases.

The first use case is for survey statisticians working with complex sampling designs.  
(I'll put the second in a separate comment.)

For starters, {rsample} could allow stratification when strata *are* constant within each group, and throw an error otherwise. That alone would already be useful to survey statisticians, even if it wouldn't cover other use cases.  
That's because for typical complex sampling survey designs, "groups" are usually completely nested within strata. So the case you're worried about (what if the stratification variable isn't constant within a group?) shouldn't usually happen.

Also in this situation, if some strata have too few groups (for ex. the user requests 5-fold CV but some strata only have 4 groups), I'd want the user to get an informative error insisting that they themselves should pool strata. The strata for survey designs are usually chosen carefully, with subject matter expertise, and it would not be meaningful for my software to pool them for me.

Finally, in this situation I'd want the software to give me a choice about whether the number of *groups* or the number of ultimate *observations* should be similar across folds. If the dataset came from a survey design in which we had sampled a certain number of groups (no matter how many observations were in each group), then CV should try to get equal numbers of groups per fold. But if instead the survey design was to keep sampling groups until a target number of final units was reached, then CV should try to get equal numbers of observations per fold.

------

[...]

PR #360, opened 2022-08-19
Add a first draft of stratification with groups

mikemahoney218:
Hi all!

Following some conversation in #317 I figured I'd post a first draft of what stratification with grouping might look like. This first draft _only_ handles `group_vfold_cv()` when `balance = "groups"`.

This version requires `strata` to be constant across all groups. If users have a non-constant variable they want to use for stratification, I think it makes sense to force them to create the strata (as I do in the example), because any automated binning implement will probably be non-optimal.

This implementation differs from `make_strata()` in that `strata` doesn't need to be quoted.

I'm not sure how you feel about putting the arguments after `...` -- because I only added `strata` and `pool`, and didn't include `breaks`, `nunique`, or `depth`, I didn't want to let people pass stratification arguments by position in case those other arguments get added later.

Addresses #317

### DIFF
--- a/.github/workflows/R-CMD-check-hard.yaml
+++ b/.github/workflows/R-CMD-check-hard.yaml
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: ubuntu-18.04,   r: 'release'}
+          - {os: ubuntu-20.04,   r: 'release'}
 
     env:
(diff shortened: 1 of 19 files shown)


hfrick (approved):
Looks great, thank you! 🎈 Some lines have gotten really long, could you break those up, please? I think we're good to go after that.

mikemahoney218:
Sounds good! I merged your pillar branch into this one to make CI work.

PR #363, opened 2022-08-28
Randomize strata

mikemahoney218:
This is a follow-up to #360 . I realized that the function was deterministic for any strata with only one group, and that it would always "short change" the last strata in the vector by giving it fewer folds as options (if `length(unique(strata)) %% v != 0`). Both of these strike me as small, but non-ideal bugs in the function; this PR changes the function so that the order of strata themselves are randomized, which I believe fixes both behaviors.

Sorry to pitch a breaking change right after we merged the first version of the feature -- I had this as a shower thought this morning and realized it wouldn't be that hard to change.

I made this reprex and I don't know that it's actually helpful at all, but here's proof that this stratified sampling works:

``` r
library(rsample)
set.seed(11)

group_table <- tibble::tibble(
  group = 1:100,
  outcome = sample(c(rep(0, 70), rep(1, 30)))
)
observation_table <- tibble::tibble(
  group = sample(1:100, 1e5, replace = TRUE),
  observation = 1:1e5
)
sample_data <- dplyr::full_join(group_table, observation_table, by = "group")

strata_rate <- purrr::map(
  1:100,
  \(x) {
    rs4 <- group_vfold_cv(sample_data, group, v = 5, strata = outcome)
    purrr::map_dbl(
      rs4$splits,
      function(x) {
        dat <- as.data.frame(x)$outcome
        mean(dat == "1")
      }
    )
  }
) |> unlist()

base_rate <- purrr::map(
  1:100,
  \(x) {
    rs4 <- group_vfold_cv(sample_data, group, v = 5)
    purrr::map_dbl(
      rs4$splits,
      function(x) {
        dat <- as.data.frame(x)$outcome
        mean(dat == "1")
      }
    )
  }
) |> unlist()

# Mean absolute error of strata proportions, versus expected
mean(abs(0.3 - strata_rate))
#> [1] 0.001477664
mean(abs(0.3 - base_rate))
#> [1] 0.01895246
```

<sup>Created on 2022-08-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/make_groups.R
+++ b/R/make_groups.R
@@ -130,12 +130,22 @@ balance_groups_strata <- function(data_ind, v, ...) {
   #
   # This means that, if nrow(keys) %% v == 0, each fold should have
   # the same number of groups from each strata
-  folds <- split_unnamed(folds, sort(keys$..strata))
+  #
+  # We randomize "keys" here so that the function is stochastic even for
+  # strata with only one group:
+  unique_strata <- unique(keys$..strata)
+  keys_order <- sample.int(length(unique_strata))
+
+  # Re-order the keys data.frame based on the reshuffled strata variable:
+  keys <- keys[
+    order(match(keys$..strata, unique_strata[keys_order])),
+  ]
+
+  # And split both folds and keys with the reordered strata vector:
+  folds <- split_unnamed(folds, keys$..strata)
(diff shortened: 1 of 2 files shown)


hfrick (approved):
Good catch, thank you! 🚿

Issue #328, opened 2022-07-01
Make default for `v` consistent between `vfold_cv()` and `group_vfold_cv()`


juliasilge:
There is a lot of nice consistency between the new grouped variants implemented in #313, #315, and #316, but the original, older grouped resampling function `group_vfold_cv()` doesn't have the same default as its sister function `vfold_cv()`. With the default of `v = NULL`, it currently returns what is also called "leave-group-out" CV.

As @mattwarkentin correctly points out in #324:

> Over time this may be confusing as users may grow to expect the `group_*` version of something to return a sampling pattern similar to the default for its non-grouped sibling.

Unfortunately, making these defaults consistent would be a breaking change, and _also_ unfortunately it's the kind of breaking change we don't have good options for to communicate with users about. On the other hand, we know that `group_vfold_cv()` isn't a heavily used function in rsample.

🎯 What is the best option to pursue in this situation?

mattwarkentin:
Related to the consistency of arguments, should the newly-added `repeats` argument for `group_vfold_cv()` also come after `v` in the arg list, as is the case for its non-grouped sibling `vfold_cv()`?

It borks the positional arg matching with the current implementation:
``` r
library(rsample)

vfold_cv(mtcars, 5, 2)
#> #  5-fold cross-validation repeated 2 times 
#> # A tibble: 10 × 3
#>    splits         id      id2  
#>    <list>         <chr>   <chr>
#>  1 <split [25/7]> Repeat1 Fold1
#>  2 <split [25/7]> Repeat1 Fold2
#>  3 <split [26/6]> Repeat1 Fold3
#>  4 <split [26/6]> Repeat1 Fold4
#>  5 <split [26/6]> Repeat1 Fold5
#>  6 <split [25/7]> Repeat2 Fold1
#>  7 <split [25/7]> Repeat2 Fold2
#>  8 <split [26/6]> Repeat2 Fold3
#>  9 <split [26/6]> Repeat2 Fold4
#> 10 <split [26/6]> Repeat2 Fold5

group_vfold_cv(mtcars, cyl, 5, 2)
#> Error in `group_vfold_cv()`:
#> ! `balance` must be a character vector, not a number.
```

mikemahoney218:
> Related to the consistency of arguments, should the newly-added `repeats` argument for `group_vfold_cv()` come after `v` in the arg list (i.e., before `balance`), as is the case for its non-grouped sibling `vfold_cv()`?

[...]

PR #364, opened 2022-09-08
Make `balance = observations` work with strata

mikemahoney218:
Here's a draft implementation of how stratification might work with `balance = observations` for `group_vfold_cv()`. 

``` r
library(rsample)
set.seed(11)

group_table <- tibble::tibble(
  group = 1:100,
  outcome = sample(c(rep(0, 89), rep(1, 11)))
)
observation_table <- tibble::tibble(
  group = sample(1:100, 1e5, replace = TRUE),
  observation = 1:1e5
)
sample_data <- dplyr::full_join(group_table, observation_table, by = "group")
group_vfold_cv(sample_data, group, v = 5, strata = outcome, balance = "observations", pool = 0.1)
#> # Group 5-fold cross-validation 
#> # A tibble: 5 × 2
#>   splits                id       
#>   <list>                <chr>    
#> 1 <split [79958/20042]> Resample1
#> 2 <split [80731/19269]> Resample2
#> 3 <split [79898/20102]> Resample3
#> 4 <split [79299/20701]> Resample4
#> 5 <split [80114/19886]> Resample5

strata_rate <- purrr::map(
  1:100,
  \(x) {
    rs4 <- group_vfold_cv(sample_data, group, v = 5, strata = outcome, balance = "observations", pool = 0.1)
    purrr::map_dbl(
      rs4$splits,
      function(x) {
        dat <- as.data.frame(x)$outcome
        mean(dat == "1")
      }
    )
  }
) |> unlist()

base_rate <- purrr::map(
  1:100,
  \(x) {
    rs4 <- group_vfold_cv(sample_data, group, v = 5, balance = "observations")
    purrr::map_dbl(
      rs4$splits,
      function(x) {
        dat <- as.data.frame(x)$outcome
        mean(dat == "1")
      }
    )
  }
) |> unlist()

# Mean absolute error of strata proportions, versus expected
mean(abs(0.10 - strata_rate))
#> [1] 0.009782044
mean(abs(0.10 - base_rate))
#> [1] 0.01529062
```

<sup>Created on 2022-09-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I'm not 100% convinced this is the right way to do things; let me know if anything here seems wonky.

Addresses #317 (I'll say "fixed" when we have balance_prop as well).

### DIFF
--- a/R/make_groups.R
+++ b/R/make_groups.R
@@ -199,9 +199,8 @@ balance_observations_helper <- function(data_split, v, target_per_fold) {
   # (to ensure that each fold gets at least one group assigned):
   freq_table <- freq_table[sample.int(nrow(freq_table)), ]
   freq_table$assignment <- NA
-  # Assign _either_ the first
-  starting_folds <- seq_len(v)[seq_len(min(nrow(freq_table), v))]
-  freq_table$assignment[starting_folds] <- starting_folds
+  # Assign the first `v` rows to folds, so that each fold has _some_ data:
+  freq_table$assignment[seq_len(v)] <- seq_len(v)
 
   # Each run of this loop assigns one "NA" assignment to a fold,
   # so we won't get caught in an endless loop here
--- a/R/make_groups.R
+++ b/R/make_groups.R
@@ -168,42 +168,76 @@ balance_observations <- function(data_ind, v, ...) {
   n_obs <- nrow(data_ind)
   target_per_fold <- 1 / v
 
-  freq_table <- vec_count(data_ind$..group, sort = "location")
-  freq_table <- freq_table[sample.int(nrow(freq_table)), ]
-  freq_table$assignment <- NA
-  freq_table$assignment[seq_len(v)] <- seq_len(v)
-
-  while (any(is.na(freq_table$assignment))) {
-    next_row <- which(is.na(freq_table$assignment))[[1]]
-    next_size <- freq_table[next_row, ]$count
-
-    group_breakdown <- freq_table %>%
-      stats::na.omit() %>%
-      dplyr::group_by(.data$assignment) %>%
-      dplyr::summarise(count = sum(.data$count), .groups = "drop") %>%
-      dplyr::mutate(prop = .data$count / n_obs,
-                    pre_error = abs(.data$prop - target_per_fold),
-                    if_added_count = .data$count + next_size,
-                    if_added_prop = .data$if_added_count / n_obs,
-                    post_error = abs(.data$if_added_prop - target_per_fold),
-                    improvement = .data$post_error - .data$pre_error)
-
-    most_improved <- which.min(group_breakdown$improvement)
-    freq_table[next_row, ]$assignment <-
-      group_breakdown[most_improved, ]$assignment
-  }
+  freq_table <- balance_observations_helper(data_ind, v, target_per_fold)
 
   collapse_groups(freq_table, data_ind, v)
 
 }
 
 balance_observations_strata <- function(data_ind, v, ...) {
-  rlang::abort(
-    c(
-      "`balance = 'observations'` has not yet been implemented for grouped resampling with stratification.",
-      i = "Consider setting `balance = 'groups'`"
-    )
+  rlang::check_dots_empty()
+
+  target_per_fold <- 1 / v
+
+  data_splits <- split_unnamed(data_ind, data_ind[["..strata"]])
+  freq_table <- purrr::map_dfr(
+    data_splits,
+    balance_observations_helper,
+    v = v,
+    target_per_fold = target_per_fold
   )
+
+  collapse_groups(freq_table, data_ind, v)
+}
+
+balance_observations_helper <- function(data_split, v, target_per_fold) {
+
+  n_obs <- nrow(data_split)
+  # Create a frequency table counting how many of each group are in the data:
+  freq_table <- vec_count(data_split$..group, sort = "location")
+  # Randomly shuffle that table, then assign the first few rows to folds
+  # (to ensure that each fold gets at least one group assigned):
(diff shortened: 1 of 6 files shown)


R/make_groups.R
```diff
@@ -168,42 +168,76 @@ balance_observations <- function(data_ind, v, ...) {
...
+  collapse_groups(freq_table, data_ind, v)
+
+}
+
+balance_observations_strata <- function(data_ind, v, ...) {
+  rlang::check_dots_empty()
+
+  target_per_fold <- 1 / v
+
+  data_splits <- split_unnamed(data_ind, data_ind[["..strata"]])
```
mikemahoney218, line 182: This is the core difference between stratification and not in this implementation: without strata, `data_ind` is broken into `v` folds which are ~balanced by the number of observations; with strata, each split of the data.frame (one per strata) is broken into `v` folds which are ~balanced, which are then combined at the end,

R/make_groups.R
```diff
@@ -168,42 +168,76 @@ balance_observations <- function(data_ind, v, ...) {
...
+      # Group by fold assignments and count data in each fold:
       dplyr::group_by(.data$assignment) %>%
       dplyr::summarise(count = sum(.data$count), .groups = "drop") %>%
-      dplyr::mutate(prop = .data$count / n_obs,
-                    pre_error = abs(.data$prop - target_per_fold),
-                    if_added_count = .data$count + next_size,
-                    if_added_prop = .data$if_added_count / n_obs,
-                    post_error = abs(.data$if_added_prop - target_per_fold),
-                    improvement = .data$post_error - .data$pre_error)
+      # Calculate...:
```
mikemahoney218, line 221: The only actual change here is that I commented the code; the functionality is the same

R/vfold.R
```diff
@@ -268,18 +268,47 @@ group_vfold_cv <- function(data, group = NULL, v = NULL, repeats = 1, balance =
 group_vfold_splits <- function(data, group, v = NULL, balance, strata = NULL, pool = 0.1) {
 
   group <- getElement(data, group)
+  max_v <- length(unique(group))
   if (!is.null(strata)) {
     strata <- getElement(data, strata)
     strata <- as.character(strata)
     strata <- make_strata(strata, pool = pool)
+
+    if (is.null(v)) {
```
mikemahoney218, line 277: This is related to #328 -- the default in `group_vfold_cv()` is to set `v` to the number of unique groups, which doesn't make sense when stratifying. This code instead sets `v` to the number of unique groups in the strata with the fewest groups, and errors if that would be less than 5; it warns when `v = NULL` about this non-standard behavior.

If we did #328 to not have `v = NULL` then a lot of this would go away.

This is a breaking change to `group_vfold_cv(balance = "groups", strata = something)`, as the number of folds returned will be different than before. The old behavior never hit CRAN and I would be shocked if it was used yet.

R/vfold.R
```diff
@@ -268,18 +268,47 @@ group_vfold_cv <- function(data, group = NULL, v = NULL, repeats = 1, balance =
...
+      # to ensure that all folds get each strata
+      max_v <- min(
+        vec_count(
+          vec_unique(
+            data.frame(group, strata)
+          )$strata
+        )$count
+      )
+      message <- c(
+        "Leaving `v = NULL` while using stratification will set `v` to the number of groups present in the least common strata."
```
hfrick, line 288:
```suggestion
        "Leaving `v = NULL` while using stratification will set `v` to the number of groups present in the least common stratum."
```

hfrick: that's the correct singular of strata in English, right?

R/vfold.R
```diff
@@ -268,18 +268,47 @@ group_vfold_cv <- function(data, group = NULL, v = NULL, repeats = 1, balance =
...
+      )
+
+      if (max_v < 5) {
+        rlang::abort(c(
+          message,
+          x = glue::glue("The least common strata only had {max_v} groups, which may not be enough for cross-validation."),
```
hfrick, line 294:
```suggestion
          x = glue::glue("The least common stratum only had {max_v} groups, which may not be enough for cross-validation."),
```

tests/testthat/test-vfold.R
```diff
@@ -227,9 +227,12 @@ test_that("grouping -- other balance methods", {
 test_that("grouping -- strata", {
   set.seed(11)
 
+  common_class <- 70
+  rare_class <- 30
```
hfrick, line 231:
```suggestion
  n_common_class <- 70
  n_rare_class <- 30
```

tests/testthat/test-vfold.R
```diff
@@ -227,9 +227,12 @@ test_that("grouping -- other balance methods", {
...
+
   group_table <- tibble(
     group = 1:100,
-    outcome = sample(c(rep(0, 70), rep(1, 30)))
+    outcome = sample(c(rep(0, common_class), rep(1, rare_class)))
```
hfrick, line 235:
```suggestion
    outcome = sample(c(rep(0, n_common_class), rep(1, n_rare_class)))
```

hfrick (approved):
Thank you! 🙌 Looks good!

I think the reason it does "slightly worse" in your reprex is because both are unstratified? The call to `group_vfold_cv()` for calculating `strata_rate` is missing the `strata` argument. When we add that, we get a more fitting result.

R/vfold.R
mikemahoney218: Yup! Whoops/

mikemahoney218:
> I think the reason it does "slightly worse" in your reprex is because both are unstratified

... :facepalm: I edited my text, that is actively embarassing :laughing:

hfrick:
How about you don't send a PR until you're past your exam and _fully_ recovered? 😄 🤝

PR #366, opened 2022-10-03
remove README.md from .Rbuildignore

EmilHvitfeldt:
This PR removes `README.md` from `.Rbuildignore` making it so it will show up on CRAN and within the Package Manager UI.


hfrick:
Thank you!

PR #367, opened 2022-10-04
Refactor stratification

mikemahoney218:
Hi all! This is a PR to simplify the `make_groups()` code, which got a bit hairy when stratification was added. This should be an "invisible" change as best as I can tell, but I think will make this easier to maintain going forward.


hfrick (approved):
Thanks, that does make it nicer! I think you want to pass on the dots to `balance_groups_*()` since you check that they are empty inside those functions. Could you have a quick look at my suggested changes?

mikemahoney218:
Looks good!

PR #368, opened 2022-10-24
use vec_detect_missing instead of vc_equal_na

EmilHvitfeldt:
> [vec_equal_na()](https://vctrs.r-lib.org/reference/vec_equal_na.html) has been renamed to [vec_detect_missing()](https://vctrs.r-lib.org/reference/missing.html) to align better with vctrs naming conventions. [vec_equal_na()](https://vctrs.r-lib.org/reference/vec_equal_na.html) will stick around for a few minor versions, but has been formally soft-deprecated ([#1672](https://github.com/r-lib/vctrs/issues/1672)).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,7 +35,7 @@ Imports:
     tibble,
     tidyr,
     tidyselect,
-    vctrs (>= 0.3.0)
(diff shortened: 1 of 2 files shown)


hfrick (approved):
Thank you!

PR #380, opened 2022-11-29
Remove overly strict test

DavisVaughan:
This isn't really testing much, and is handled by the other test anyways

We are going to change `slice(df)` to drop all rows rather than keep them all. This is mainly of theoretical interest, and it makes the internal dplyr code a little simpler. https://github.com/tidyverse/dplyr/pull/6573

rsample seems to be the only package affected by this, because of an overly strict test I added while checking dplyr compatibility .


hfrick (approved):
Thanks!

PR #384, opened 2022-12-01
Turn accessor functions into generics

hfrick:
This PR turns the functions to access data sets from `rsplit` objects into generics:

- `training()`
- `testing()`
- `analysis()`
- `assessment()`

~This PR does not change current behavior but I was wondering if `training()` and `testing()` should work on all `rsplit` objects or only those that are returned by `initial_split()`?~ -> #394


topepo (approved):

hfrick:
The answer to my question on accessor functions is yes, resulting in this issue: #394

That means this is ready to merge, but I'll wait until after the next release so that all things related to the validation split rework can sit in dev for a while.

topepo (approved):

PR #391, opened 2022-12-06
remove docs for old dplyr compat

### LINKED ISSUE #388: Remove help page on dplyr compatibility
[`R/compat-dplyr.R`](https://rsample.tidymodels.org/dev/reference/rsample-dplyr.html) talks about version-specific behaviour for dplyr versions older than `v1.0.0` but by now we require dplyr v1.0.0 so we can remove that.

hfrick:
closes  #388

I think this is a leftover from  #306 where the compat code for old (< v1.0.0) dplyr versions was removed.


DavisVaughan:
i dont think we should just nuke these docs. i think they provide valuable rationale for why the operations return tibbles vs rset objects in the current version of rsample/dplyr. they probably just need a refresh where we remove sentences where we talk about old dplyr/rsample, and remove columns from the tables that talk about them

hfrick:
Okay, thanks! The info about what happens with different versions was what I was focussed on, but I agree, the info on what happens with current versions of rsample and dplyr is still good to know. I've brought back that content and updated it accordingly.

R/compat-dplyr.R
```diff
@@ -1,106 +1,62 @@
 #' Compatibility with dplyr
 #'
 #' @description
-#' rsample should be fully compatible with dplyr 1.0.0.
+#' This page lays out the compatibility between rsample and dplyr. The `rset`
+#' objects from rsample are a sepeciifc subclass of tibbles, hence standard
+#' dyplyr operations like joins as well row or column modifications work.
```
DavisVaughan, line 6:
```suggestion
#' objects from rsample are a specific subclass of tibbles, hence standard
#' dplyr operations like joins as well row or column modifications work.
```

Maybe do one more full pass for typos?

DavisVaughan (approved):

R/compat-dplyr.R
hfrick: thanks!

Issue #394, opened 2022-12-06
Differentiate between initial split and splits in resamples in the accessor functions `training()`/ `testing()` / `analysis()`/`assessment()`


hfrick:
TMwR calls the elements of the initial split "training" and "testing" set and the elements of the a resample "analysis" and "assessment" set. 
![](https://www.tmwr.org/premade/resampling.svg)

In rsample, both of those types of splits are `rsplit` objects and all of `training()`/ `testing()` / `analysis()`/`assessment()` work on all `rsplit` objects. We should be more consistent with that conceptual difference and make `training()`/ `testing()` work only on initial splits and `analysis()`/`assessment()` only on splits in resamples.

``` r
library(rsample)

split_init <- initial_split(mtcars)
cars_train <- training(split_init)
cars_analysis <- analysis(split_init) # this should error

folds <- vfold_cv(mtcars, v = 3)
fold_1 <- folds$splits[[1]]
analysis_1 <- analysis(fold_1)
train_1 <- training(fold_1) # this should error
```

<sup>Created on 2022-12-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hfrick:
Follow-up question to ponder, possibly it's own issue: Do we want any of these functions to work on `rset` objects? 

This is motivated by some of those sets only containing one split anyway (the ones for a validation split/set and the `apparent` set). For example, for a validation set, it feels sorta natural to want to be able to access the one analysis (aka training) data from the `rset`. 

For those sets containing multiple splits, `analysis()` and `assessment()` could gain an `index` argument similar to `get_rsplit(x, index)`. `training()` could return the `$data` of the splits.

PR #404, opened 2022-12-13
Add `validation_set()` for 3-way split approach

hfrick:
This is part 2 of introducing a 3-way split approach to the validation split: This PR adds a function `validation_set()` to make an `rset` from the result of `initial_validation_split()` and `group_initial_validation_split()`.

The accessor methods are currently as follows, with `val_split` being the class of the split objects inside the `rset`:

 accessor / object           | `initial_validation_split` | `val_split`
---            | ---                        | ---        
`training()`   | training                   | training   
`validation()` | validation                 | validation 
`testing()`    | testing                    | error      
`analysis()`   | error                      | training   
`assessment()` | error                      | validation 

For other types of `rsplit` objects inside of an `rset`, `training()` and `testing()` will error (see #394). I've made them work here because the split object isn't actually resampled (all the sampling happens in `initial_validation_split()`) but I'd like a second opinion here! We can also make `training()` and `validation()` error on `val_split` objects.

Similarly, do we want accessor functions for the `rset` returned by `validation_set()`? For other resamples, these do not exist but for this case, it sorta feels like they should? Or more precisely: should it work because it's just one split in the set? This would also apply to the other sets with only one split: `validation_split`, `validation_time_split`, `group_validation_split`, and `apparent`. I can also open a separate issue on this if we want to ponder this further.


R/validation_set.R
```diff
@@ -0,0 +1,77 @@
+#' Create a Validation Split for Tuning
+#'
+#' @param x An object of class `initial_validation_split`, such as resulting
```
hfrick, line 3: What's a good name for this argument?

R/validation_set.R
```diff
@@ -0,0 +1,77 @@
...
+#' set.seed(1353)
+#' car_split <- initial_validation_split(mtcars)
+#' car_set <- validation_set(car_split)
+validation_set <- function(x, ...) {
+  rlang::check_dots_empty()
+
+  train_and_val <- rbind(
+    training(x),
+    validation(x)
+  )
```
hfrick, line 22: This changes the order of rows compared to the initial data set going into `initial_validation_split()` - should it be the same order of rows?

R/validation_set.R
topepo: `split`?

R/validation_set.R
topepo: I don't think that it is a big deal.

R/validation_set.R
```diff
@@ -0,0 +1,77 @@
...
+#' @export
+validation.val_split <- function(x, ...) {
+  assessment(x, ...)
+}
+
+#' @rdname validation_set
+#' @export
+testing.val_split <- function(x, ...) {
+  rlang::abort(
+    "The testing data is not part of the validation set.", # TODO change error message?
```
topepo, line 74:
```suggestion
    "The testing data is not part of the validation set object.", 
```

topepo (approved):

PR #410, opened 2023-01-30
use testthat::skip_if_not_installed()

MichaelChirico:
This leads to better messages in the test log (as compared to e.g. 'rlang::is_installed("withr") is not TRUE (2)')


hfrick:
Thanks! This looks good but I'll fix the test failures unrelated to this so that we can see more easily that the checks are passing on this before I merge.

PR #414, opened 2023-02-08
dplyr 1.1.0 and multiple matches in joins

hfrick:
This PR adapts rsample to dplyr 1.1.0: multiple matches in joins now cause warnings unless we set `multiple = "all"`.

That argument `multiple` is new in v1.1.0 so I've also bumped the minimum version up. In other packages, prior to the CRAN release of dplyr v1.1.0, we only used that argument conditionally, e.g.
```
if (utils::packageVersion("dplyr") >= "1.0.99.9000") {
    pred <- full_join(param_key, pred, by = "group", multiple = "all")
  } else {
    pred <- full_join(param_key, pred, by = "group")
  }
```
Happy to change it to this pattern if we think we don't want to depend on v1.1.0 just yet!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,7 +22,7 @@ Depends:
     R (>= 3.4)
 Imports: 
     cli,
-    dplyr (>= 1.0.0),
+    dplyr (>= 1.1.0),
     furrr,
     generics,
     glue,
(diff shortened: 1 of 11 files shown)


EmilHvitfeldt (commented):
Looking great!

Issue #158, opened 2020-06-01
feature request - manual split creation


DanOvando:
I'm exploring using tidymodels for my predictive modeling workflows and loving it so far. I am running into one problem though. I nearly always want to do my initial split by something like year or country, but have many nested observations within those splits. I can't figure out a clean way of creating these splits using rsample. To use the gapminder dataset, suppose I want to train a model using only data from before the year 2000, and test the model on all data points greater than or equal to the year 2000. 

I see how this could be done combining initial_time_split with nesting, but the resulting object would not (I assume) work if passed as the `split` argument of `tune::last_fit`, since each resample would be a nested tibble. 

My current strategy is to arrange the data such that my training and testing splits are cleanly separated, find the split point in the data, and then use that as the prop argument, to initial_time_split, but that feels pretty rough to me, and makes me nervous with very large datasets, where I'm not sure how exact rsample is being in the splitting (e.g. if the correct split is 0.961234, will rsample be that precise?). 

Would it be possible to add something in to `initial_split` like `training_split`, where `training_split` could be a logical vector saying TRUE if so, FALSE if not? That way, users could make the definition of the testing and training splits as complex as needed, and simply pass the resulting logical vector to `initial_split`. 

``` r
library(tidyverse)
library(tidymodels)
#> ── Attaching packages ──────────────────────────────── tidymodels 0.1.0 ──
#> ✓ broom     0.5.6      ✓ rsample   0.0.6 
#> ✓ dials     0.0.6      ✓ tune      0.1.0 
#> ✓ infer     0.5.1      ✓ workflows 0.1.1 
#> ✓ parsnip   0.1.1      ✓ yardstick 0.0.6 
#> ✓ recipes   0.1.12
#> ── Conflicts ─────────────────────────────────── tidymodels_conflicts() ──
#> x scales::discard() masks purrr::discard()
#> x dplyr::filter()   masks stats::filter()
#> x recipes::fixed()  masks stringr::fixed()
#> x dplyr::lag()      masks stats::lag()
#> x dials::margin()   masks ggplot2::margin()
#> x yardstick::spec() masks readr::spec()
#> x recipes::step()   masks stats::step()
library(gapminder)

gapminder <- gapminder %>% 
  arrange(year)

split_prop <- (last(which(gapminder$year <= 2000))) / nrow(gapminder)

test <- initial_time_split(gapminder, prop = split_prop)

# gapminder works in 5 year intervals, so max year should be 1997 in training set
[...]

PR #422, opened 2023-03-18
Adding an example for the make_splits function

AngelFelizR:
The example was taken from the [Max Kuhn](https://github.com/topepo) issue answer.

https://github.com/tidymodels/rsample/issues/158


hfrick:
Thank you for the PR @AngelFelizR ! Max's example works well as an illustration in that issue but isn't quite what we'd put in an example on the help page: We like to avoid `\dontrun{}` if possible, and I'd like to avoid taking a dependency on gapminder just for this example. Would you like to take another stab at this with simple (possibly artificial) data which highlights the usage of `make_splits()` for a list and maybe a data frame, without any additional dplyr manipulations?

PR #428, opened 2023-04-20
update snaps

hfrick:


hfrick:
a bit too early, these changes are only in dev versions that I had installed

PR #429, opened 2023-04-21
Enforce empty dots when not used

### LINKED ISSUE #427: Enforce empty dots when not used
Prompted by https://lsinks.github.io/posts/2023-04-10-tidymodels/tidymodels_tutorial.html#resampling-via-rsample

> I did notice that the typo “stata” didn’t kick up any sort of error. In fact, you can include any number of named parameters that don’t exist, and you won’t get an error. Positional matching shouldn’t apply if you are using named parameters, but for what it is worth, initial_split(fraud, prop = 0.75, mouse = is_fraud) and initial_split(fraud, mouse = is_fraud, prop = 0.75 ) both execute without complaint. And they both produce stratified samples, which is weird. Don’t rely on this and do check that your splits and folds are properly stratified.

This is part 2 of #381 but I think it's better placed in its own issue.

hfrick:
closes #427

Generally, this PR checks for empty dots via `rlang::check_dots_empty()` whenever they are documented as "not used".

For the `tidy()` methods, it was documented that the dots are not used. However, that was only true for some of the methods, see e.g., the tests for `tidy.vfold_cv()` which passed `unique_id` through the dots. Now all general methods (for `rsplit`, `rset`, `nested_cv`) have an explicit `unique_ind` arg and the dots have to be empty. For `vfold_cv`, there is no such arg because each observation only gets assigned to one fold (per repeat) and thus `unique_ind` should not make a difference for the tidy output. 

The dots are generally documented via the rlang tag. Because that (also) states that the dots must be empty, the dots for a generic are documented with "Not currently used." only (since, for the generic, they do not need to be empty). 

Methods for base-R generics like `print()` and `as.integer()` have _unchecked_ dots documented as "Not currently used".

Revdeps: `"1a6d222f-b434-4844-972f-805b5fb6f0b7"`
This breaks 2 revdeps which are both now fixed in their dev version:
- butcher https://github.com/tidymodels/butcher/pull/259
- healthyR.ts https://github.com/spsanderson/healthyR.ts/pull/439

Do we typically add the revdep files to the PR? rsample already has a `revdep/` folder under version control, and I'm wondering if I'm adding this here or leaving them untouched until release time.


DavisVaughan (approved):
Seems reasonable to me. Nice that it caught some typos in the tests

PR #436, opened 2023-06-08
Expand the grouping columns for percentile and t confidence intervals

topepo:
For use with `tune_result` objects, this PR allows `int_pctl()` and `int_t()` to `group_by()` more columns than just `term`. 

This expands the list to `"term"`, `".config"`, `".iter"`, and `".eval_time"`.

Since this is a developer-facing change, it is undocumented (in Rd file) and I left it out of NEWS.md

### DIFF
--- a/R/bootci.R
+++ b/R/bootci.R
@@ -72,13 +72,14 @@ check_tidy <- function(x, std_col = FALSE) {
       x <-
         dplyr::select(x, dplyr::all_of(grp_cols), estimate, id,
                       tidyselect::one_of(std_candidates)) %>%
-        mutate(id = (id == "Apparent")) %>%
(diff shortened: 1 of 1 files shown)


R/bootci.R
```diff
@@ -255,7 +268,6 @@ int_pctl <- function(.data, ...) {
 #' @export
 #' @rdname int_pctl
 int_pctl.bootstraps <- function(.data, statistics, alpha = 0.05, ...) {
-  check_dots_empty()
```
hfrick, line 258: Why does this get removed?

hfrick (commented):
I'm a little reluctant of defining the grouping so far away from where it's used, i.e. in here rsample when it's used in tune. However, if there is a lot to be gained in tune by this, then I'm okay with it

Issue #443, opened 2023-08-03
row binding data frames with Surv objects


topepo:
`validation_set()` uses `rbind()` to [concatenate](https://github.com/tidymodels/rsample/blob/46de181d5900a6be182509b192bf51c89985075a/R/validation_set.R#L20:L23) the training and validation sets. 

This fails for `Surv` objects (but `dplyr::bind_rows()` does the right thing): 

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(survival)

lung$suv_obj <- Surv(lung$time, lung$status)
lung_1 <- head(lung)
lung_2 <- tail(lung)


r_bound <- rbind(lung_1, lung_2)
inherits(r_bound$suv_obj, "Surv")
#> [1] FALSE
class(r_bound$suv_obj)
#> [1] "matrix" "array"

dplyr_bound <- bind_rows(lung_1, lung_2)
inherits(dplyr_bound$suv_obj, "Surv")
#> [1] TRUE
class(dplyr_bound$suv_obj)
#> [1] "Surv"
```

<sup>Created on 2023-08-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

topepo:
There are a lot of other places that this will occur when the `Surv` object is in a basic R data frame. 

For example, the partitioning functions use base R subscripting: 

```r
x$data[sort(x$train_id), , drop = FALSE]
```

and that drops the class. These can be changed to: 

```r
  ind <- sort(x$train_id)
  dplyr::slice(x$data, ind)
```

and it solves the issue. 

We would have to do this change in a lot of places: 

- `{training,testing,validation}.initial_validation_split`
-  `as.data.frame.rsplit`

so it would be a system-wide change. For some methods, the latter uses column subscripting at the same time so I would also use `dplyr::select()` to keep the `Surv` objects intact.

topepo:
I think I know what to do about [these lines](https://github.com/tidymodels/rsample/blob/46de181d5900a6be182509b192bf51c89985075a/R/rsplit.R#L112:L114) but will hold off before making changes there. 

```r
permuted_col <-
   x$data[as.integer(x, data = data, ...), x$col_id, drop = FALSE]
x$data[, x$col_id] <- permuted_col
```

topepo:
[...]

PR #444, opened 2023-08-03
Change row indexing/subsetting to preserve Surv objects

topepo:
See #443 for details. 

Keep as draft PR so that we can decide how to proceed and if the row name changes are acceptable.


R/rsplit.R
```diff
@@ -114,7 +114,8 @@ as.data.frame.rsplit <-
       x$data[, x$col_id] <- permuted_col
       return(x$data)
     }
-    x$data[as.integer(x, data = data, ...), , drop = FALSE]
+    row_ind <- as.integer(x, data = data, ...)
+    dplyr::slice(x$data, row_ind)
```
simonpcouch, line 118: `as.data.frame.rsplit()` sees a lot of traffic inside of tune and ends up being a non-negligible portion of time elapsed in `fit_resamples()` and friends. With this change, I see:

``` r
bench::mark(
   old = x$data[as.integer(x, data = data, ...), , drop = FALSE],
   new = dplyr::slice(x$data, as.integer(x, data = data, ...))
)
#> # A tibble: 2 × 13
#>   expression      min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
#>   <bch:expr> <bch:tm> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 old           452µs  485µs     2009.     778KB     45.6   749    17
#> 2 new           591µs  643µs     1506.     823KB     40.6   594    16
#> # ℹ 5 more variables: total_time <bch:tm>, result <list>,
#> #   memory <list>, time <list>, gc <list>
```

Not a dealbreaker, but worth considering.

simonpcouch (commented):
Generally in support of this!

R/rsplit.R
EmilHvitfeldt: If we are worried about speed, we could use `vctrs::vec_slice()`

``` r
library(survival)

lung$suv_obj <- Surv(lung$time, lung$status)

bench::mark(
  base = lung[1:4, , drop = FALSE],
  dplyr = dplyr::slice(lung, 1:4),
  vctrs = vctrs::vec_slice(lung, 1:4)
)
#> # A tibble: 3 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 base        34.44µs   36.8µs    26262.    20.5KB     26.3
#> 2 dplyr      235.18µs  244.2µs     4012.    3.67MB     20.9
#> 3 vctrs        9.18µs     10µs    96547.        0B     29.0
```

PR #449, opened 2023-08-17
Soft-deprecate `validation_split()`

### LINKED ISSUE #350: Deprecate `validation_split()`
This is a fairly confusing name because it doesn't produce a _split_ object like `initial_split()` does. It produces a single resample object. Maybe it should be named `validation_set()`, or simply `validation()`.

I think we could provide an alias for it with a better name and immediately deprecate the old name with `lifecycle::deprecate_warn()` (still works, but users get a warning once every session).

hfrick:
closes #350

This updates the help page for `validation_split()` and its variants, adds the deprecation warning, and updates the tests accordingly. 

It also updates the vignette that mentions it. We don't use it in extratests, so no changes there. Max was working on updating tidymodels.org and TMwR.


EmilHvitfeldt (approved):
Looking good!

Issue #478, opened 2024-05-16
Change class for grouped mc split


hfrick:
The function is `group_mc_split()` but the class is `grouped_mc_split` whereas for vfold_cv and bootstraps they are both `group_*`. They should all be `group_`.

https://github.com/tidymodels/rsample/blob/a976356791cc13e3fb6a5672b8b98d43931fb24d/R/mc.R#L242

PR #483, opened 2024-05-21
Potato/inner/calibration split 

hfrick:
First pass at adding those split methods with a few examples. I'll leave (self-)review comments for discussion.

TODO (to follow in separate PRs)
- (initial) validation split
- bootstrap
- slider variants

Not planned
- LOO: tune doesn't want to do this (and there is no intention to do so)
- nested resamples: tune currently doesn't do this so we can wait with that one too
- permutations: there is no assessment set so it doesn't work in tune either
- manual resamples: If someone manually crafted their resampling, it most likely is nothing of what rsample already contains so it does not make sense to me to fall back on any of our already implemented option. 

Additional notes on things to still add and where:
- Whether or not the inner split is stratified on the outcome should be addressed in tune.
- Similarly, what to do when there is no sensible inner split option should probably also be in tune, rsample will error as usual for now.


R/inner_split.R
```diff
@@ -0,0 +1,161 @@
+#' Inner split of the analysis set for fitting a post-processor
+#' 
+#' @param x An `rsplit` object.
+#' @param split_args A list of arguments to be used for the inner split.
+#' @param ... Not currently used.
+#' @return An `rsplit` object.
+#' @keywords internal
+#' @export
+inner_split <- function(x, ...) {
```
hfrick, line 9: Are we happy with that name? We can also go back and change it later (like container -> tailor).

R/inner_split.R
```diff
@@ -0,0 +1,161 @@
+#' Inner split of the analysis set for fitting a post-processor
+#' 
+#' @param x An `rsplit` object.
+#' @param split_args A list of arguments to be used for the inner split.
+#' @param ... Not currently used.
+#' @return An `rsplit` object.
```
hfrick, line 6: Maybe this is slightly awkward but the choice is for easy of handling in tune, given that we don't anticipate much of a use case for it outside of tune.

R/inner_split.R
```diff
@@ -0,0 +1,161 @@
...
+
+#' @rdname inner_split
+#' @export
+inner_split.vfold_split <- function(x, split_args, ...) {
+  check_dots_empty() 
+
+  analysis_set <- analysis(x)
+  
+  # TODO should this be done outside of rsample, 
+  # in workflows or tune?
```
hfrick, line 71: The idea here is that you can hand `inner_split()` the set of arguments used to create the `rset` or `initial_split` and `inner_split()` doing the necessary translation if `prop` is still unspecified.

That makes is easier to handle via the `get_split_args()` helper and easier to document what can be in `split_args`: the default is "the same arguments used to create the `rset` or `initial_split`" and then only need to document exceptions like this where `prop` is also possible.

R/inner_split.R
```diff
@@ -0,0 +1,161 @@
...
+# clustering -------------------------------------------------------------
+
+#' @rdname inner_split
+#' @export
+inner_split.clustering_split <- function(x, split_args, ...) {
+  check_dots_empty() 
+
+  analysis_set <- analysis(x)
+  
+  # TODO: reduce the number of clusters by 1 in tune?
```
hfrick, line 129: Given that the basic idea of `clustering_cv()` is to use _one cluster_ as the assessment set, I would reduce `v` by one for the inner split, so that the cluster left out for the inner split is more likely to be similar to one of the original clusters. If we use the same `v`, the inner clustering is likely to break up the `v-1` clusters in this (outer) analysis set. 

I would put that into tune though, not here.

tests/testthat/test-inner_split.R
```diff
@@ -0,0 +1,165 @@
+
+# mc ---------------------------------------------------------------------
+
+test_that("mc_split", {
+  set.seed(11)
+  r_set <- mc_cv(warpbreaks)
+  split_args <- get_split_args(r_set)
```
hfrick, line 7: I would expect this or something similar to be the pattern in tune. If so, we can export this helper function, maybe as `.get_split_args()`.

R/inner_split.R
```diff
@@ -0,0 +1,161 @@
...
+    analysis_set, 
+    prop = split_args$prop, 
+    times = 1,
+    strata = split_args$strata, 
+    breaks = split_args$breaks, 
+    pool = split_args$pool
+  )
+  split_inner <- split_inner$splits[[1]]
+
+  class_inner <- paste0(class(x)[1], "_inner")
```
EmilHvitfeldt, line 32: not to be pedantic, but wouldn't `class_inner` by definition be `"mc_split_inner"`? same for other `class_inner`. While i appreciate the same code being used across, I think we could just note the class directly

R/inner_split.R
```diff
@@ -0,0 +1,161 @@
...
+  analysis_set <- analysis(x)
+  
+  split_inner <- mc_splits(
+    analysis_set, 
+    prop = split_args$prop, 
+    times = 1,
+    strata = split_args$strata, 
+    breaks = split_args$breaks, 
+    pool = split_args$pool
+  )
```
EmilHvitfeldt, line 29: this current design assumes that we remember to pass in all the arguments to `mc_splits()` correctly.

If we envision that we have to do this more times, we could do a little metaprogramming to have the arguments passed through.

```r
  split_args$times <- 1
  split_inner <- rlang::inject(mc_splits(
    analysis_set, 
    !!!split_args
  ))
```

or 

```r
cl <- rlang::call2("ms_splits", .ns = "rsample",data = analysis_set)
split_args$times <- 1
cl <- rlang::call_modify(cl, !!!split_args)
split_inner <- rlang::eval_tidy(cl)
```

EmilHvitfeldt (approved):

R/inner_split.R
simonpcouch: I guess what I'm suggesting is that the answer to this comment is "yes" and we'd call `inner_split()` from tune/workflows as:

```
  split_args <- get_split_args(r_set)
  r_split <- get_rsplit(r_set, 1)

  # ...where prop is extracted from `add_tailer(prop)`
  isplit <- inner_split(r_split, split_args, prop = .8)
```

rather than:

```
  split_args <- get_split_args(r_set)
  r_split <- get_rsplit(r_set, 1)
  # since the r_set doesn't know where the supplied data comes from (or whether
  # it was generated by `initial_split()` or friends)
  split_args$prop <- .8

  isplit <- inner_split(r_split, split_args)
```

tests/testthat/test-inner_split.R
simonpcouch: Yes, this would be awesome to have access to :)

simonpcouch: Naming-wise, noting the existence of `get_rsplit()` might make `get_rsplit_args()` (or prefixed with dot) a friend

simonpcouch (commented):
Need to go AFK for a couple hours so going to go ahead and send off comments from my first pass through! I think this is in a really good spot already.

R/inner_split.R
hfrick: That is a good question. I was going with trying to make the function signature the same for ease in tune - but we need to special case _somewhere_. I'll open an issue for this and we can iterate between rsample and tune a little to feel this out.

hfrick: #487

R/inner_split.R
hfrick: Do you mean that in reference to this particular method or for all the methods? 

Since I had just come across #478, I opted for constructing the class rather than writing it out manually here to make sure it would always stay in sync with the class of the input `x`. In terms of readability, I would say that the class of that one is fairly easy to see from the S3 dispatch.

tests/testthat/test-inner_split.R
simonpcouch: I just discovered `tune::pull_rset_attributes()`. Possibly a similar idea?

hfrick: The tune function pulls a bigger set of attributes with the ones we're interested in here nested in the output of the tune function. I'm open to trying to make that one function but am also fine with them co-existing.

Re name: That pair of names is a nice idea. Would it be confusing if it didn't always return the exact arguments of the `rsplit`? That can happen for e.g. vfold_cv where it also return the `repeats` argument which is only relevant for the `rset` but not the `rsplit` inside. The object for an `initial_validation_split` is not an `rsplit` in the first place since it's a `threeway_split`. Is that too pedantic? 😄

simonpcouch: Yup, looks like `tune::pull_rset_attributes` does provide a good bit more info. Since we were able to make do in https://github.com/tidymodels/tune/pull/894 with that just function, no pressure from me to export or otherwise mobilize further here. :)

R/inner_split.R
EmilHvitfeldt: what i meant, is that each `paste0(class(x)[1], "_inner")` in this file could be swapped with a `mc_split_inner`, `apparent_split_inner`, etc etc as they are called inside s3 methods, on the object that drives the s3 dispatch

PR #100, opened 2019-07-11
Stratification var

jyuu:
Eliminated these tests

`expect_error(mc_cv(iris, strata = 2))`
`expect_error(vfold_cv(iris, strata = 2))`
  
So user can specify strata by index of column as it is behavior familiar from tidyselect

Closes #95

PR #486, opened 2024-05-22
Allow explicit specification of `strata = NULL` for grouped resampling

### LINKED ISSUE #485: Grouped resampling breaks with non-missing `strata = NULL`
The default for `strata` is `NULL` but specifying it explicitly breaks `check_grouped_strata()`.

``` r
library(rsample)

set.seed(11)
rs1 <- group_vfold_cv(warpbreaks, "tension")

rs1 <- group_vfold_cv(warpbreaks, "tension", strata = NULL)
#> Error in .subset2(x, i, exact = exact): attempt to select less than one element in get1index
```

<sup>Created on 2024-05-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

hfrick:
closes  #485

``` r
library(rsample)

set.seed(11)
rs1 <- group_vfold_cv(warpbreaks, "tension")

rs1 <- group_vfold_cv(warpbreaks, "tension", strata = NULL)
```

<sup>Created on 2024-05-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 * `vfold_cv()` now utilizes the `breaks` argument correctly for repeated cross-validation (@ZWael, #471).
 
+* Grouped resampling functions now work with an explicit `strata = NULL` instead of strata being either a name or missing (#485).
+
 ## Breaking changes
 
 * The class of grouped MC splits is now `group_mc_split` instead of `grouped_mc_split`, aligning it with the other grouped splits (#478).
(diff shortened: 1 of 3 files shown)


simonpcouch (approved):
This does the trick!

Is there a reason why this change is in `check_grouped_strata()` rather than just changing the condition in:

https://github.com/tidymodels/rsample/blob/c82c2a22d9297c50a47203db38d98864b5020653/R/vfold.R#L220-L222

to `if (!is.null(strata))`?

hfrick:
yeah... just doing that switch breaks being able to supply an unquote name for `strata`. I think it would need to be evaluated for this condition but unevaluated for `check_grouped_strata()` and that looked like the worse effort/effect ratio than what I did there. 

I have since noticed that I was not alone in going down this route: https://github.com/tidymodels/rsample/pull/100

If we're touching this, we should do it across rsample.

PR #489, opened 2024-05-23
New subclasses and `inner_split()` methods for validation sets

hfrick:
This PR adds `inner_split()` methods for the `rset`s coming from `validation_set()`. Those `rsets` can result from the `group` and `time` variants of `initial_validation_split()`. 

To be able to tell where they are coming from, this PR also adds subclasses for the `group` and `time` variant to the output of `validation_set()`: the `rset` objects and the `rsplit` objects within those `rset`s.

I'm looking for a second opinion on the naming of the `rsplit` subclasses for the `time` variant. 

The interfaces and the original (and deprecated) approach via `validation_split()` put the `time` neither as a prefix nor a suffix but rather _inside_ the more general name: `initial_time_split()`, `initial_validation_time_split()`, `validation_time_split()`, and `val_time_split`.

I don't think we are going to touch the naming of the interfaces but we could touch the naming convention for the `rsplit` class. I've broken with that particular pattern of placing `time` and instead aligned with the pattern for `group`, i.e., used it as a prefix. Could be good, could be bad, could actually not matter much. 

The only place I think it'll matter is here and I'm happy to add a method for `val_time_split` to support the usage of the deprecated `validation_time_split()` if we think it's necessary. 
Edit: [this search](https://github.com/search?q=org%3Atidymodels%20val_time_split&type=code) suggests that we don't dispatch on that class elsewhere so it should only matter here.


R/inner_split.R
```diff
@@ -18,6 +18,11 @@
 #' `split_args$balance = NULL`.
 #' * For `boot_split` and `group_boot_split` objects, it will ignore
 #' `split_args$times`.
+#' * For `val_split`, `group_val_split`, and `time_val_split` objects, it will
```
simonpcouch, line 21: This has a nice ring to it.

> I've broken with that particular pattern of placing `time` and instead aligned with the pattern for `group`, i.e., used it as a prefix.

Let's move forward with this! As principled a move as any, I think.

R/inner_split.R
```diff
@@ -18,6 +18,11 @@
 #' `split_args$balance = NULL`.
 #' * For `boot_split` and `group_boot_split` objects, it will ignore
 #' `split_args$times`.
+#' * For `val_split`, `group_val_split`, and `time_val_split` objects, it will 
+#' interpret a length-2 `split_args$prop` as a ratio between the training and
+#' validation sets and split into inner analsysi and inner assessment set in
```
simonpcouch, line 23:
```suggestion
#' validation sets and split into inner analysis and inner assessment set in 
```

re: "inner assessment", this language is a bit :thinkies:. Is this a potato set or is this something else since we're supposedly doing validation?

R/misc.R
```diff
@@ -302,7 +302,14 @@ non_random_classes <- c(
...
   all_attributes <- attributes(rset)
-  args <- names(formals(all_attributes$class[[1]]))
+  function_used_to_create <- switch(
+    all_attributes$class[[1]],
+    "validation_set" = "initial_validation_split",
+    "group_validation_set" = "group_initial_validation_split",
+    "time_validation_set" = "initial_validation_time_split",
+    all_attributes$class[[1]]
+  )
+  args <- names(formals(function_used_to_create))
```
simonpcouch, line 312: Just thinking about our usage of the other helper in tune. There's no conflict here because these kinds of splits would never be tune input, right? h.b. `last_fit()`?

Maybe the answer is to just refactor https://github.com/tidymodels/tune/pull/894/commits/3ba3c49794341d676bd49fd7b150fa2a38d651db with `.get_split_args` so we don't have to think about this anymore.

R/inner_split.R
```diff
@@ -166,6 +172,83 @@ inner_split.group_boot_split <- function(x, split_args, ...) {
...
+#' @rdname inner_split
+#' @export
+inner_split.val_split <- function(x, split_args, ...) {
+  check_dots_empty() 
+
+  analysis_set <- analysis(x)
+
+  if (length(split_args$prop) == 2) {
+    # keep ratio between training and validation as ratio between
+    # inner analysis and inner assessment
```
simonpcouch, line 186: Same thinkies re: inner assessment vs potato.

R/inner_split.R
```diff
@@ -166,6 +172,83 @@ inner_split.group_boot_split <- function(x, split_args, ...) {
...
+    split_args$prop <- split_args$prop[[1]] / sum(split_args$prop)
+  } else {
+    split_args$prop <- split_args$prop[[1]]
+  }
+  split_inner <- initial_time_split(
+    analysis_set,
+    prop = split_args$prop
+  )
```
simonpcouch, line 244: Had to pause for a moment and trace back through arguments for this one. Feels strange to pass around `split_args` only for `prop`...

So, `initial_time_split()` can also take `lag` here, but that would never be in `split_args` because it's not an argument to `initial_validation_time_split()`/`validation_set()`. Do we ever imagine an extension to those latter two where `split_args` would end up with a `lag` in it that ought to be plumbed through? May be worth a code comment in that case, though may also be worth me tracking down a second cup of coffee.

simonpcouch (approved):
yup yup yup yup

> I'm happy to add a method for `val_time_split` to support the usage of the deprecated `validation_time_split()` if we think it's necessary.

I'd say we not do so. :)

R/inner_split.R
hfrick: yes, the inner assessment set is the potato set. I wasn't sure where we were with the naming and I'm missing a way to differentiate what I call outer and inner analysis set and was therefore fully leaning into inner/outer naming here also also applying it to the assessment/potato set. 
You're not the only one though: https://github.com/tidymodels/rsample/issues/490

R/misc.R
hfrick: > these kinds of splits would never be tune input, right?

I'm not entirely sure which splits you mean here, so here's the summary which hopefully (also) answers your question :) 

The result of `initial_validation_split()` (and friends) can go into `last_fit()`. It does not go into any tuning function like `tune_grid()`, the `rset` for those functions comes from `validation_set()`.

The objects from `initial_validation_split()` are of class `initial_validation_split` and `three_way_split`.
The objects from `validation_set()` are `rsets` with splits inside. Those splits are `rsplit`s (aka binary splits instead of the earlier 3-way split) with a `val_split` subclass.

R/inner_split.R
hfrick: Thanks for asking this, I'll leave a comment! We've decided against a `lag` argument for `initial_validation_split()` -- the slider functions are much more principled in how they handle lags and if we add functionality for an initial validation split with lags, it would be through slider variants of `initial_validation_split()` rather than the current `initial_validation_time_split()`.

R/misc.R
simonpcouch: > The result of initial_validation_split() (and friends) can go into last_fit(). It does not go into any tuning function like tune_grid(), the rset for those functions comes from validation_set().

Heard. :) Should have spoken a bit more to this—internally, `last_fit()` [dispatches to]( https://github.com/tidymodels/tune/blob/544d54160e6c1d5c05622afcf2c3ec276dc4303a/R/last_fit.R#L203-L211) the machinery underlying `fit_resamples()`, which uses the machinery underlying `tune_grid()`. So `last_fit()` itself will end up pulling off attributes from the `rset` [manually constructed](https://github.com/tidymodels/tune/blob/544d54160e6c1d5c05622afcf2c3ec276dc4303a/R/last_fit.R#L193-L197) internally from the result of `initial_validation_split()` and friends.

hfrick: Oooh, so that manual rset sent from `last_fit()` to, ultimately, `tune_grid()` needs to keep the/more attributes if we want `tune_grid()` to be able to use `inner_split()` - that's what you mean, right?

Why don't I leave https://github.com/tidymodels/rsample/pull/495 open for now and if don't end up using it, I'll clean up at the end.

simonpcouch: Sounds good. :)

R/inner_split.R
hfrick: Ended up obscuring that by sticking to the meta-programming pattern 🏄

PR #495, opened 2024-05-24
Don't export `.get_split_args()`

hfrick:
until we need to and 

> Since we were able to make do in https://github.com/tidymodels/tune/pull/894 with that just function, no pressure from me to export or otherwise mobilize further here.

https://github.com/tidymodels/rsample/pull/483#discussion_r1611702826


simonpcouch:
I did just complicate this story in https://github.com/tidymodels/rsample/pull/489#discussion_r1613507422. Will defer to you here.🤔

hfrick:
Summary of discussion under https://github.com/tidymodels/rsample/pull/489#discussion_r1613507422
We'll leave this PR open for now and see if we end up using this helper in tune after all. If not, we'll merge this.

simonpcouch:
tune now does indeed use this helper, so I will go ahead and close.🐬

PR #501, opened 2024-07-19
Update to roxygen2 version 3.7.2

hfrick:


hfrick:
CI failure on R 3.6 is okay, we don't support that anymore (see also #500)

PR #516, opened 2024-08-15
cli updates for `bootci.R`

Dpananos:
Addresses #504 

I've tried to use pluralization where possible.

Note that running `testthat::test_file('tests/testthat/test-bootci.R')` results in one failure

```
[ FAIL 1 | WARN 0 | SKIP 0 | PASS 52 ]

── Failure (test-bootci.R:138:3): Upper & lower confidence interval does not contain NA ──
Snapshot of code has changed:
old[4:7] vs new[4:7]
    Warning:
    Recommend at least 1000 non-missing bootstrap resamples for term `mean`.
    Error in `pctl_single()`:
-   ! All statistics have missing values..
+   ! All statistics have missing values.

* Run testthat::snapshot_accept('bootci') to accept the change.
* Run testthat::snapshot_review('bootci') to interactively review the change.
[ FAIL 1 | WARN 0 | SKIP 0 | PASS 52 ]
```

one of the tests expects two periods.  I wasn't sure if that was a typo or not.  I can switch back to using 2 periods or I can edit the test.  Let me know!


Dpananos:
I added back the extra period to get the tests to pass.

Dpananos:
Need to format with {`.arg`} as mentioned here https://github.com/tidymodels/rsample/pull/518#discussion_r1718700413

hfrick:
I'm looking at this PR now, but we definitely only need one period at the end! 😆

R/bootci.R
```diff
@@ -131,14 +131,8 @@ check_num_resamples <- function(x, B = 1000) {
...
-    msg <-
-      paste0(
-        "Recommend at least ", B, " non-missing bootstrap resamples for ",
-        ifelse(length(terms) > 1, "terms: ", "term "),
-        paste0(terms, collapse = ", "),
-        "."
-      )
-    rlang::warn(msg)
+    cli::cli_warn(paste0("Recommend at least {B} non-missing bootstrap ",
+                         "resamples for {cli::qty(terms)} term{?s} {terms}."))
```
hfrick, line 133: Let's simplify this a little further: we can skip the `paste0()` part and instead use `x$term` directly in `cli_warn()` for the pluralization. As an illustration:

``` r
terms <- letters[1:3]; cli::cli_warn("for {terms} term{?s}")
#> Warning: for a, b, and c terms

terms <- letters[1]; cli::cli_warn("for {terms} term{?s}")
#> Warning: for a term

terms <- letters[1:2]; cli::cli_warn("for {terms} term{?s}")
#> Warning: for a and b terms
```

<sup>Created on 2024-08-15 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Dpananos:
@hfrick I think this is ready for review now.  I've updated the arguments to use the `{.arg}` formatting, and I've caught some more `rlang::aborts` I initially missed.

I also remove the extra period.

hfrick (commented):
Thanks for putting in the styling changes already! Could you also take a look at the one code comment here and not namespace `cli::cli_abort()` here (because we already import it)?

R/bootci.R
Dpananos: OK!  I think I understood the assignment.  `cli::cli_warn` is not imported, but `cli::cli_abort` is.  Are we good namespace `cli_warn` in this case, or should we import it as well?

Dpananos:
Checks still might fail.  I've rolled back changes I made to the snapshot since those were in error.

hfrick:
Thanks @Dpananos for the PR! I've fiddled a little with the pluralization for the `term{?s}` bit and now we're set!

PR #517, opened 2024-08-15
Uses cli_abort in R/caret.R

### LINKED ISSUE #505: Use cli errors for `R/caret.R`

#499 can serve as an example.

Dpananos:
closes #505


Dpananos:
Need to format with `{.arg}` as mentioned here https://github.com/tidymodels/rsample/pull/518#discussion_r1718700413

Dpananos:
This should be ready for review

hfrick:
@Dpananos Thanks a bunch for the multiple PRs! I'm just dropping in quickly to say that I'm OOO until the week after next but will get to this after my return.

PR #518, opened 2024-08-15
Use cli erros in rsplit.R Fixes #512

### LINKED ISSUE #512: Use cli errors in `R/rsplit.R`

#499 can serve as an example.

JamesHWade:


R/rsplit.R
```diff
@@ -1,20 +1,20 @@
 rsplit <- function(data, in_id, out_id) {
   if (!is.data.frame(data) & !is.matrix(data)) {
-    rlang::abort("`data` must be a data frame.")
+    cli::cli_abort("`data` must be a data frame.")
```
hfrick, line 3: Could you please add styling for `data`? It's an argument and it's wrapped in `{.arg}` like in https://github.com/tidymodels/rsample/pull/499/files, it will get the appropriate styling.

The same applies to the other arguments referenced in the error messages.

R/rsplit.R
```diff
@@ -88,13 +88,13 @@ as.data.frame.rsplit <-
            data = "analysis",
            ...) {
     if (!is.null(row.names)) {
-      rlang::warn(paste0(
+      cli::cli_warn(paste0(
```
hfrick, line 91: For cli, we also don't need the `paste0()` anymore, could you update this accordingly as well, please?

hfrick (commented):
Awesome! I left two comments directly on the code and have one high-level one: We already import `cli_abort()` so there is no need to namespace it as `cli::cli_abort()`.

PR #519, opened 2024-08-15
Fixes #475 - Documentation edits-decluttering the top of the refs page

### LINKED ISSUE #475: Declutter start of reference page
Not all methods may need to show up here, it's currently pretty crowded.

image

laurabrianna:
Fixes #475 - moved args into details section for first three functions to keep references page doc decluttered

### DIFF
--- a/R/initial_split.R
+++ b/R/initial_split.R
@@ -6,7 +6,9 @@
 #'  `group_initial_split` creates splits of the data based
 #'  on some grouping variable, so that all data in a "group" is assigned to
 #'  the same split.
-#'  `training` and `testing` are used to extract the resulting data.
+#'
+#' @details `training` and `testing` are used to extract the resulting data.
+#'
 #' @template strata_details
 #' @inheritParams vfold_cv
 #' @inheritParams make_strata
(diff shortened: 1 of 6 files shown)


R/initial_split.R
```diff
@@ -6,7 +6,9 @@
 #'  `group_initial_split` creates splits of the data based
 #'  on some grouping variable, so that all data in a "group" is assigned to
 #'  the same split.
-#'  `training` and `testing` are used to extract the resulting data.
+#'
+#' @details `training` and `testing` are used to extract the resulting data.
```
hfrick, line 10: I think it's fine to keep those where they were so not moving them to `@details` but it would be lovely if they gained a `()` at the end, so e.g. `training()` instead of `training` - then they are easier to identify as functions!

R/validation_set.R
```diff
@@ -1,5 +1,7 @@
 #' Create a Validation Split for Tuning
 #'
+#' `validation_set` creates a the validation split for model tuning.
```
hfrick, line 3: This would also look great with a set of `()` at the end

hfrick (commented):
Thanks so much for the contribution! As we just chatted, I'll turn the clutter on the reference page into a pkgdown issue but there are some small but nice styling changes possible here anyway, are you up for fixing those?

PR #520, opened 2024-08-15
Updated example for nested_cv()

### LINKED ISSUE #479: Update example for `nested_cv()`
The illustration of why using the bootstrap for the outer resampling is a bad idea -- it's not in the example (anymore): the row in question is not in the outer analysis set.
Also, using the accessor functions `analysis()` and `assessment()` (and `get_rsplit()`) would be nicer than `as.data.frame()`.

seb09:
Replaced as.data.frame() by the rsample accessor functions and changed the car name - Fixes #479

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 * Started moving error messages to cli (#499, #502).
 
+* Fixed example for `nested_cv()` (@seb09, #520).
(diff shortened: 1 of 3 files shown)


hfrick (approved):
Thank you for fixing that! 🙌

PR #521, opened 2024-08-15
use `fun()` instead of `fun` across docs, fixes #383

### LINKED ISSUE #383: Use `function()` instead of `function` in documentation
This follows the style guide: https://style.tidyverse.org/documentation.html#r-code

brshallo:
Fixes #383 

Mainly impacts older docs that did not follow [style guide](https://style.tidyverse.org/documentation.html#r-co)

### DIFF
--- a/R/caret.R
+++ b/R/caret.R
@@ -4,10 +4,10 @@
 #'  \pkg{rsample} and \pkg{caret}.
 #'
 #' @param object An `rset` object. Currently,
-#'  `nested_cv` is not supported.
-#' @return `rsample2caret` returns a list that mimics the
+#'  `nested_cv()` is not supported.
+#' @return `rsample2caret()` returns a list that mimics the
 #'  `index` and `indexOut` elements of a
-#'  `trainControl` object. `caret2rsample` returns an
+#'  `trainControl` object. `caret2rsample()` returns an
 #'  `rset` object of the appropriate class.
 #' @export
 rsample2caret <- function(object, data = c("analysis", "assessment")) {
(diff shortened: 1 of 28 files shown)


topepo:
Weird question... is this part of our tidy dev day? I want to make sure that it doesn't overlap with something that someone is actively working on.  No big deal either way, but good to know.

R/caret.R
```diff
@@ -4,10 +4,10 @@
 #'  \pkg{rsample} and \pkg{caret}.
 #'
 #' @param object An `rset` object. Currently,
-#'  `nested_cv` is not supported.
-#' @return `rsample2caret` returns a list that mimics the
+#'  `nested_cv()` is not supported.
```
topepo, line 7: Instead of backticks, can you use brackets so cross-linking happens? For example: 

```suggestion
#'  [nested_cv()] is not supported.
```

R/caret.R
```diff
@@ -4,10 +4,10 @@
...
+#' @return `rsample2caret()` returns a list that mimics the
 #'  `index` and `indexOut` elements of a
-#'  `trainControl` object. `caret2rsample` returns an
+#'  `trainControl` object. `caret2rsample()` returns an
```
topepo, line 10: For external functions, you can namespace them

```suggestion
#'  `[caret::trainControl()]` object. [caret2rsample()] returns an
```

hfrick: We'll trigger a CRAN note when cross-referencing functions from packages we don't depend on so let's keep references to such packages (like caret here) in backticks instead of switching to the brackets.

hfrick:
@brshallo Thanks a bunch for the PR! I'm just dropping in quickly to say that I'm OOO until the week after next but will get to this after my return.

PR #524, opened 2024-08-15
Mark `rolling_origin()` as superseded

### LINKED ISSUE #448: Mark `rolling_origin()` as superseded
in favor of `sliding_window()`, `sliding_index()` and `sliding_period()`

nmercadeb:
Closes #448

### DIFF
--- a/vignettes/Common_Patterns.Rmd
+++ b/vignettes/Common_Patterns.Rmd
@@ -223,11 +223,3 @@ sliding_period(Chicago, date, "year") %>%
   head(2)
 ```
 
-All of these functions produce analysis sets of the same size, with the start and end of the analysis set "sliding" down your data frame. If you'd rather have your analysis set get progressively larger, so that you're predicting new data based upon a growing set of older observations, you can use the `rolling_origin()` function:
-
-```{r}
-rolling_origin(Chicago) %>%
-  head(2)
-```
-
-Note that all of these time-based resampling functions are deterministic: unlike the rest of the package, running these functions repeatedly under different random seeds will always return the same results. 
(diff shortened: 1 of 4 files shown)


vignettes/Common_Patterns.Rmd
```diff
@@ -223,11 +223,3 @@ sliding_period(Chicago, date, "year") %>%
   head(2)
 ```
```
mikemahoney218, line 225:
```suggestion
All of these functions produce analysis sets of the same size, with the start and end of the analysis set "sliding" down your data frame. If you'd rather have your analysis set get progressively larger, so that you're predicting new data based upon a growing set of older observations, you can use the `sliding_window()` function with `lookback = -Inf`:

```{r}
sliding_window(Chicago, lookback = Inf) %>%
  head(2)
```

Note that all of these time-based resampling functions are deterministic: unlike the rest of the package, running these functions repeatedly under different random seeds will always return the same results. 
```

mikemahoney218: ^ I think the documentation here is still pretty useful, because as discussed in #448 I think we need to highlight how to approach this common use case. It might also be useful to highlight `skip` here for creating larger initial training sets in this context.

hfrick (commented):
Thank you for your contribution! Could you also add the corresponding badge to the documentation as laid out here: https://lifecycle.r-lib.org/articles/communicate.html#supersede-a-function

hfrick:
@nmercadeb Thank you for the edits! I'm just dropping in quickly to say that I'm OOO until the week after next but will get to this after my return.

PR #526, opened 2024-08-15
for slide.R and changing all the rlang::abort to cli_abort

### LINKED ISSUE #513: Use cli errors in `R/slide.R`

#499 can serve as an example.

PriKalra:
changed all rlang::abort to cli_abort. There were lot of paste0 rlang abort, they have been changed to proper syntax for cli_abort.
fixes #513

### DIFF
--- a/R/slide.R
+++ b/R/slide.R
@@ -499,7 +499,7 @@ check_lookback <- function(x) {
 
 check_assess <- function(x, arg) {
   if (vctrs::vec_size(x) != 1L) {
-    cli_abort("{.arg arg} must have size 1.You have provided {.arg {arg}}")
+    cli_abort("{.arg {arg}} must have size 1.")
   }
 
   if (identical(x, Inf)) {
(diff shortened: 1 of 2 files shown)


hfrick:
@PriKalra Thanks a bunch for this PR and the others! I'm just dropping in quickly to say that I'm OOO until the week after next but will get to this after my return.

R/slide.R
```diff
@@ -216,7 +216,7 @@ sliding_window <- function(data,
   rlang::check_dots_empty()
 
   if (!is.data.frame(data)) {
-    cli_abort("{.arg data} must be a {.cls data frame}.")
+    cli_abort("{.arg data} must be a data frame.")
```
hfrick, line 219: The reason for this change is a fairly subtle one: "data frame" refers to data frames in general, both of the base-R `data.frame` flavour and of the tibble flavour. We could use `{.cls data.frame}` here because tibbles are also `data.frame`s but the error might suggest tibbles are not allowed if a user doesn't know this technical detail. Therefore I'm sticking with the general "data frame" here.

``` r
data.frame(x = 10) |> class()
#> [1] "data.frame"

tibble::tibble(x = 10) |> class()
#> [1] "tbl_df"     "tbl"        "data.frame"
```

<sup>Created on 2024-09-09 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

R/slide.R
```diff
@@ -499,19 +499,19 @@ check_lookback <- function(x) {
 
 check_assess <- function(x, arg) {
   if (vctrs::vec_size(x) != 1L) {
-    cli_abort("{.arg arg} must have size 1.You have provided {.arg {arg}}")
```
hfrick: It is generally a good idea to add info on what the argument was in addition to what it is supposed to be! 👍 I'm only removing this here because we'll be doing this a different way in the very near future.

PR #527, opened 2024-08-15
Prevent LOO through vfold_cv()

### LINKED ISSUE #440: Prevent LOO through `vfold_cv()`
We purposefully don't (fully) support LOO cross-validation in tidymodels: we have `loo_cv()` in rsample but it's not supported in tune.

However, you can create LOO cross-validation by doing:
```
vfold_cv(df,  v = nrow(df))
```
which then does not get caught in tune. Let's make `vfold_cv()` error for this case, with a reference to `loo_cv()`.

``` r
library(tidymodels)

loo_folds <- loo_cv(mtcars)
v_folds <- vfold_cv(mtcars, v = nrow(mtcars))

wflow <- workflow(mpg ~ ., linear_reg())

fit_resamples(wflow, resamples = loo_folds)
#> Error in `check_rset()` at tune/R/tune_grid.R:334:2:
#> ! Leave-one-out cross-validation is not currently supported with tune.

fit_resamples(wflow, resamples = v_folds)
#> → A | warning: A correlation computation is required, but the inputs are size zero or one and the standard deviation cannot be computed. `NA` will be returned.
#> There were issues with some computations   A: x1
#> There were issues with some computations   A: x32
#> 
#> # Resampling results
#> # 32-fold cross-validation 
#> # A tibble: 32 × 4
#>    splits         id     .metrics         .notes          
#>    <list>         <chr>  <list>           <list>          
#>  1 <split [31/1]> Fold01 <tibble [2 × 4]> <tibble [1 × 3]>
#>  2 <split [31/1]> Fold02 <tibble [2 × 4]> <tibble [1 × 3]>
#>  3 <split [31/1]> Fold03 <tibble [2 × 4]> <tibble [1 × 3]>
#>  4 <split [31/1]> Fold04 <tibble [2 × 4]> <tibble [1 × 3]>
#>  5 <split [31/1]> Fold05 <tibble [2 × 4]> <tibble [1 × 3]>
#>  6 <split [31/1]> Fold06 <tibble [2 × 4]> <tibble [1 × 3]>
#>  7 <split [31/1]> Fold07 <tibble [2 × 4]> <tibble [1 × 3]>
#>  8 <split [31/1]> Fold08 <tibble [2 × 4]> <tibble [1 × 3]>
#>  9 <split [31/1]> Fold09 <tibble [2 × 4]> <tibble [1 × 3]>
#> 10 <split [31/1]> Fold10 <tibble [2 × 4]> <tibble [1 × 3]>
#> # ℹ 22 more rows
#> 
#> There were issues with some computations:
#> 
#>   - Warning(s) x32: A correlation computation is required, but the inputs are size ze...
#> 
#> Run `show_notes(.Last.tune.result)` for more information.
```

<sup>Created on 2023-07-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

seb09:
vfold_cv() throws an error, if used for leave-one-out cross-validation, refering to loo_cv() instead. Fixes #440

### DIFF
--- a/R/vfold.R
+++ b/R/vfold.R
@@ -12,7 +12,9 @@
 #' @template strata_details
 #' @inheritParams make_strata
 #' @param data A data frame.
-#' @param v The number of partitions of the data set.
+#' @param v The number of partitions of the data set. Should be an integer
+#' smaller than `nrow(data)`. If you want to create a split for a leave-one-out
+#' cross-validation (`v = nrow(data)`), please use [loo_cv()] instead.
 #' @param repeats The number of times to repeat the V-fold partitioning.
 #' @param strata A variable in `data` (single character or name) used to conduct
 #'  stratified sampling. When not `NULL`, each resample is created within the
@@ -74,17 +76,20 @@ vfold_cv <- function(data, v = 10, repeats = 1,
   strata_check(strata, data)
   check_repeats(repeats)
 
+  if (isTRUE(v == nrow(data))) {
+    rlang::abort(c(
+      "Leave-one-out cross-validation is not supported by `vfold_cv()`.",
+      x = "You set `v` to `nrow(data)`, which would result in a leave-one-out cross-validation.",
+      i = "Use `loo_cv()` in this case."
+    ))
+  }
+
   if (repeats == 1) {
     split_objs <- vfold_splits(
       data = data, v = v,
       strata = strata, breaks = breaks, pool = pool
     )
   } else {
-    if (v == nrow(data)) {
-      rlang::abort(
-        glue::glue("Repeated resampling when `v` is {v} would create identical resamples")
-      )
-    }
     for (i in 1:repeats) {
       tmp <- vfold_splits(data = data, v = v, strata = strata, breaks = breaks ,pool = pool)
       tmp$id2 <- tmp$id
--- a/R/vfold.R
+++ b/R/vfold.R
@@ -124,10 +124,10 @@ vfold_cv <- function(data, v = 10, repeats = 1,
 }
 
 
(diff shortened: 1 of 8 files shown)


hfrick:
@seb09 Thanks a bunch for the PR! I'm just dropping in quickly to say that I'm OOO until the week after next but will get to this after my return.

seb09:
Enjoy your time off 🙂

R/vfold.R
```diff
@@ -74,17 +76,20 @@ vfold_cv <- function(data, v = 10, repeats = 1,
   strata_check(strata, data)
   check_repeats(repeats)
 
+  if (isTRUE(v == nrow(data))) {
```
hfrick: nice defensive `isTRUE()` 👍

R/vfold.R
```diff
@@ -12,7 +12,9 @@
 #' @template strata_details
 #' @inheritParams make_strata
 #' @param data A data frame.
-#' @param v The number of partitions of the data set.
+#' @param v The number of partitions of the data set. Should be an integer
+#' smaller than `nrow(data)`. If you want to create a split for a leave-one-out
+#' cross-validation (`v = nrow(data)`), please use [loo_cv()] instead.
```
hfrick: We also require a few other things of `v`, e.g. to be larger than 1 and don't mention it all here. We could move it to the Details section but I think it's okay to let users bump into the error for the edgecases so that we can keep their focus on more high-level aspects.

hfrick (commented):
@seb09  rsample already has a function to check that `v` is something sensible, `check_v()`, so I've moved your check there. This means we now catch it in `vfold_cv()` and `clustering_cv()`.

PR #531, opened 2024-08-29
Rsample cli errors update for #510

### LINKED ISSUE #510: Use cli errors in `R/permutations.R`, `R/reg_intervals.R`

#499 can serve as an example.

PriKalra:
Have added some creative ways of throwing cli errors recently. Thank you for looking into the code and suggesting- alongside the patience! :D 
Let me know if these changes make sense in the big picture!

fixed #510


hfrick:
Thanks so much! You've clearly learned more styling options since the previous PR, well done! I've generally "toned it down" a little bit to keep it more like the rest of error messages in rsample and elsewhere. The little arrow produced by `">"` is nice detail though that I've kept!

Issue #458, opened 2023-10-31
Use rlang type-checkers


hfrick:

Dpananos:
Is this still needed?  Happy to lean in here

PR #547, opened 2024-09-19
Check strata input

hfrick:
Doing the thing!

- rlang type checker for #458
- thread the call through for  #541

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -105,9 +105,6 @@ strata_check <- function(strata, data, call = caller_env()) {
     if (inherits(data[, strata], "Surv")) {
       cli_abort("{.arg strata} cannot be a {.cls Surv} object. Use the time or event variable directly.")
     }
-    if (!(strata %in% names(data))) {
-      cli_abort("{strata} is not in {.arg data}.")
-    }
   }
   invisible(NULL)
 }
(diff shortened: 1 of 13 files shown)


R/misc.R
```diff
@@ -98,16 +98,17 @@ add_class <- function(x, cls) {
   x
 }
 
-strata_check <- function(strata, data) {
+check_strata <- function(strata, data, call = caller_env()) {
```
EmilHvitfeldt, line 101: love the rename!

R/misc.R
```diff
@@ -98,16 +98,17 @@ add_class <- function(x, cls) {
...
-    if (!is.character(strata) | length(strata) != 1) {
-      cli_abort("{.arg strata} should be a single name or character value.")
-    }
     if (inherits(data[, strata], "Surv")) {
-      cli_abort("{.arg strata} cannot be a {.cls Surv} object. Use the time or event variable directly.")
-    }
-    if (!(strata %in% names(data))) {
-      cli_abort("{strata} is not in {.arg data}.")
+      cli_abort(c(
+        "{.arg strata} cannot be a {.cls Surv} object.",
```
EmilHvitfeldt, line 107: Technically `strata` is a character or name. but it is used to refer to column in the data set. So I think `.field` would be better here as it refers to the value

```suggestion
        "{.field strata} cannot be a {.cls Surv} object.",
```

EmilHvitfeldt (approved):
Love it 🔥

R/misc.R
hfrick: Good point, thank you!

PR #551, opened 2024-09-25
Check `prop` argument

hfrick:
This PR reworks how the `prop` argument to various functions is checked.

Several of the functions had a written-out check block in either the user-facing function or the underlying `*_split()` function. The functions for grouped resampling were checking `prop` deep down the call chain with a `check_prop()`.

I've decided to place the check in the user-facing functions and reworked it a little to benefit from the "friendly object type" in the standard checkers from rlang and only  do the boundaries in a custom way.

### DIFF
--- a/R/make_groups.R
+++ b/R/make_groups.R
@@ -231,7 +231,6 @@ balance_observations_helper <- function(data_split, v, target_per_fold) {
 
 balance_prop <- function(prop, data_ind, v, replace = FALSE, strata = NULL, ...) {
   rlang::check_dots_empty()
-  check_prop(prop, replace)
 
   # This is the core difference between stratification and not:
   #
@@ -289,20 +288,6 @@ balance_prop_helper <- function(prop, data_ind, v, replace) {
   ) %>%
     list_rbind()
 }
-
-check_prop <- function(prop, replace) {
-  acceptable_prop <- is.numeric(prop)
-  acceptable_prop <- acceptable_prop &&
-    ((prop <= 1 && replace) || (prop < 1 && !replace))
-  acceptable_prop <- acceptable_prop && prop > 0
-  if (!acceptable_prop) {
-    cli_abort(
-      "{.arg prop} must be a number between 0 and 1.",
-      call = rlang::caller_env()
-    )
-  }
-}
-
 
 collapse_groups <- function(freq_table, data_ind, v) {
   data_ind <- dplyr::left_join(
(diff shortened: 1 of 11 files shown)


R/make_groups.R
```diff
@@ -290,20 +289,6 @@ balance_prop_helper <- function(prop, data_ind, v, replace) {
     list_rbind()
 }
 
-check_prop <- function(prop, replace) {
-  acceptable_prop <- is.numeric(prop)
-  acceptable_prop <- acceptable_prop &&
-    ((prop <= 1 && replace) || (prop < 1 && !replace))
```
hfrick, line 296: This distinction between what the upper boundary can be is removed because we are using it only in `group_boot_splits()`:
https://github.com/tidymodels/rsample/blob/fe24aaa9f880960e5ef2829ffaa52439b0bf71b7/R/boot.R#L264-L265
We know where this particular `prop = 1` is coming from and that it's okay in this context so the new `check_prop()` can just check user input.

simonpcouch: saaaaaaaatisfying

R/misc.R
```diff
@@ -98,6 +98,17 @@ add_class <- function(x, cls) {
...
 }
 
+check_prop <- function(prop, call = caller_env()) {
+  check_number_decimal(prop, call = call)
+  if (!(prop > 0)) {
+    cli_abort("{.arg prop} must be greater than 0.", call = call)
+  }
+  if (!(prop < 1)) {
+    cli_abort("{.arg prop} must be less than 1.", call = call)
+  }
```
simonpcouch, line 108: Wish there was some way to mark `check_number_decimal(min)` and `max` as exclusive but here we are.

tests/testthat/test-mc.R
```diff
@@ -72,6 +72,9 @@ test_that("strata", {
 
 
 test_that("bad args", {
+  expect_snapshot(error = TRUE, {
+    mc_cv(mtcars, prop = 1)
+  })
```
simonpcouch, line 77:
```suggestion
  expect_snapshot(error = TRUE, {
    mc_cv(mtcars, prop = -1)
  })
  expect_snapshot(error = TRUE, {
    mc_cv(mtcars, prop = 1)
  })
```

...as we haven't checked the "greater than 0" message.

simonpcouch (approved):
Upgrade! Up! Grade!

R/misc.R
hfrick: Same 🫠

PR #614, opened 2026-09-17
fixes #610 with cli::cli_abort used instead of rlang::abort

### LINKED ISSUE #610: `rolling_origin()` passes a numeric row-count into `rlang::abort()`'s `class` argument instead of the message
In `R/rolling_origin.R`:

```r
if (n < initial + assess) {
  rlang::abort(
    "There should be at least ",
    initial + assess,
    " rows in `data`."
  )
}
```

`rlang::abort()`'s signature is `abort(message, class = NULL, ...)`, so the second positional argument (`initial + assess`, a number) is bound to `class` rather than being concatenated into the message. This either errors (since `class` must be a character vector) or silently produces a malformed/incomplete error message, rather than telling the user how many rows are required.

reprex:
``` r
library(rsample)
rolling_origin(mtcars[1:3, ], initial = 5, assess = 1)
#> Error in `rlang::abort()`:
#> ! `class` must be a character vector, not the number 6.
```

<sup>Created on 2026-08-31 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Suggested fix: switch to `cli_abort()`, matching the rest of the package's `cli`-based messaging:

```r
cli_abort("There should be at least {initial + assess} rows in {.arg data}.")
```

agmurray:
changed 3 instances of `rlang::abort()` to `cli::cli_abort` and used glue `{ }` style braces


EmilHvitfeldt (approved):
Wonderful!

PR #615, opened 2026-09-17
fixes #609

### LINKED ISSUE #609: `t_single()` length check uses `&&` instead of `||`, silently accepting mismatched `std_err`
In `R/bootci.R`, `t_single()` validates that `stats`, `std_err`, and `is_orig` have matching lengths:

```r
if (length(stats) != length(std_err) && length(stats) != length(is_orig)) {
  function_args <- c('stats', 'std_err', 'is_orig')
  cli_abort("{.arg {function_args}} should have the same length.")
}
```

Because `&&` requires *both* mismatches to be true simultaneously, a `std_err` whose length differs from `stats` (but where `is_orig` happens to match) passes validation silently. This can produce incorrect or garbled t-interval calculations downstream instead of raising the intended error.

Suggested fix: change `&&` to `||` so a length mismatch in either `std_err` or `is_orig` triggers the abort:

```r
if (length(stats) != length(std_err) || length(stats) != length(is_orig)) {
```

A regression test with a `std_err` of the wrong length would help catch this.

agmurray:
Change `&&` to `||` in `t_single()` and made a new test as suggested in the issue.


EmilHvitfeldt (approved):
rockstar!

